PR #64, opened 2015-06-18
Enable Nuclide Server to listen on connections over both IPv4 and IPv6

jazzdan:
Currently when trying to run Nuclide server on an OS that does not support IPv6 (in this case an old version of Linux) the server will fail to start with this error message:

```
Error: listen EAFNOSUPPORT
```

Some digging online indicated this was likely related to the OS not supporting IPv6. I removed the second argument as shown in this patch and then the server started fine. I was also able to connect from Nuclide no problem.

I *believe* this change will allow connections over both IPv4 and IPv6 on the specified port. I'm not too familiar with Node's implementation of listen() but I know there are differences between Linux, other Unixes and Windows on how listen() is handled when it comes to IPv4 and IPv6. This patch may be insufficient to allow connections over both IPv4 and IPv6 in those environments, but I don't have any Windows/Other Unixes available to test with at the moment.

### DIFF
--- a/pkg/nuclide/server/lib/NuclideServer.js
+++ b/pkg/nuclide/server/lib/NuclideServer.js
@@ -303,7 +303,7 @@ class NuclideServer {
         this._webServer.removeAllListeners();
         reject(e);
       });
-      this._webServer.listen(this._port, '::');
+      this._webServer.listen(this._port);
     });
   }
 


jazzdan:
Is there anything I can do to help this review along?

jssln:
Hey @jazzdan, can you check whether https://github.com/facebook/nuclide/commit/decbc0b4eb2e1d52af1ad9f1e0ea88b85195748a addressed your problem?

whscullin:
@jssln @jazzdan That seems to be a client side fix, the issue is with the server.

cbosuna:
It sounds like the behavior of listen changed in https://github.com/joyent/node/issues/9195.  Older node versions would only bind to ipv4 if '::' was not specified.  Let me test this pull request on a few different platforms.  If it works it will be merged.

jazzdan:
Thanks everyone! @cbosuna let me know if there is anything I can do to help test.

jazzdan:
I can confirm that this is still a problem with the latest version of the server.

cbosuna:
This does in fact break ipv6.  It seems the correct solution is to start 2 servers one on each address family.

cbosuna:
Wait, I think something screwy was going on on my box.  This may be ok.  Still testing.

cbosuna:
This does work on node 0.12.X.  It binds to both ipv4 and ipv6.  We will pull this in.  Thanks!

PR #94, opened 2015-06-24
Use mv instead of fs.rename so that cross-device renames do not fail with EXDEV

### LINKED ISSUE #88: Renaming temporary file fails when tmp dir is on different partition
As discovered in #83, babel-cache can fail if the temporary files it generates are generated on a different partition than their final destination. Sample error output on Ubuntu (as pasted from [a comment](https://github.com/facebook/nuclide/issues/83#issuecomment-114670597)):

```
TASK(981): uncaughtException: { [Error: EXDEV: cross-device link not permitted, rename '/tmp/nuclide-node-transpiler115523-981-1iiz309' -> '~/.atom/packages/nuclide-quick-open/node_modules/nuclide-client/node_modules/nuclide-server/node_modules/nuclide-node-transpiler/babel-cache/564aa7208191ab5835e8b76ad7973252d24e0b92/d922337f311d15c072e4c8da176392ac689a7726']
  errno: -18,
  code: 'EXDEV',
  syscall: 'rename',
  path: '/tmp/nuclide-node-transpiler115523-981-1iiz309',
  dest: '~/.atom/packages/nuclide-quick-open/node_modules/nuclide-client/node_modules/nuclide-server/node_modules/nuclide-node-transpiler/babel-cache/564aa7208191ab5835e8b76ad7973252d24e0b92/d922337f311d15c072e4c8da176392ac689a7726' }
```

mtorromeo:
This adds a new dependency on the mv module.
Fixes #88

### DIFF
--- a/pkg/nuclide/node-transpiler/lib/babel-cache.js
+++ b/pkg/nuclide/node-transpiler/lib/babel-cache.js
@@ -15,6 +15,7 @@ var crypto = require('crypto');
 var fs = require('fs');
 var path = require('path');
 var temp = require('temp').track();
+var mv = require('mv');
 var cacheDir = createCacheDir();
 
 /**
@@ -92,7 +93,7 @@ function createOrFetchFromCache(sourceCode, filePath) {
         return;
       }
 
-      fs.rename(info.path, transpiledFile);
+      mv(info.path, transpiledFile, {mkdirp: true});
     });
   });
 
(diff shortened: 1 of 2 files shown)


ssorallen:
Thanks for the PR. We are going to check out the `mv` NPM package and make sure it will work for everyone.

PR #102, opened 2015-06-25
Allow the flow linter to run under the source.js.jsx scope

etcinit:
Currently, the flow linter only runs on the `source.js` grammar scope. However, multiple packages that provide JSX syntax highlighting use the `source.js.jsx` scope:

- https://github.com/orktes/atom-react
- https://github.com/subtleGradient/language-javascript-jsx
- https://atom.io/packages/language-babel

There are [other scopes](https://github.com/AtomLinter/linter-eslint/blob/master/lib/linter-eslint.coffee#L36) that the linter could support but the JSX one seemed like the most important.

### DIFF
--- a/pkg/nuclide/flow/lib/FlowLinter.js
+++ b/pkg/nuclide/flow/lib/FlowLinter.js
@@ -78,7 +78,7 @@ function processDiagnostics(diagnostics: Array<Object>, targetFile: string) {
 }
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


nmote:
This looks good to me; I've imported it for internal review.

nmote:
We've merged this. It will show up next time we export commits. Thank you for contributing!

PR #109, opened 2015-06-27
cleared up confusing error message inside nuclide-diff-view

numso:
When trying to diff any files in my project, I'd get: 

```
Cannot open diff view for file: /Users/dallin/code/kualico/cm/server/resources/workflow/types/types.js
Error: git
```

Which made me think nuclide wasn't able to find my git binary. This change makes that error message a little clearer: 

```
Cannot open diff view for file: /Users/dallin/code/kualico/cm/server/resources/workflow/types/types.js
Error: Diff view only supports hg repositories right now: git
```

### DIFF
--- a/pkg/nuclide/diff-view/lib/DiffViewModel.js
+++ b/pkg/nuclide/diff-view/lib/DiffViewModel.js
@@ -38,7 +38,7 @@ class DiffViewModel {
     }
     var repository = await atom.project.repositoryForDirectory(rootDirectory);
     if (!repository || repository.getType() !== 'hg') {
-      throw new Error('Diff view only supports hg repositories right now: ' + repository && repository.getType());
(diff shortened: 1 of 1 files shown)


ssorallen:
Thanks for the fix. We will get it merged.

jssln:
@numso Could you make a small tweak to the message? It would be a bit clearer to say:
`Error: Diff view only supports hg repositories right now: **found** git`

audrummer15:
Excellent.

numso:
@jssln That makes a lot more sense. It's been done.

piotr-cz:
Also, could change [package description](https://atom.io/packages/nuclide-diff-view):

_Provides Diff Viewer agnostic to the source control type of a project_

to something like
_Provides Diff View to hg repositories
Planned fatures: GIT support_

PR #114, opened 2015-06-29
minor issues

aledista:


jssln:
Thanks!

pkg/nuclide/server/scripts/utils.py
```diff
@@ -46,7 +46,7 @@ def http_get(host, port, method, url, key_file=None, cert_file=None, timeout=1):
             return ret
         else:
             return None
-    except:
+    except Exception:
```
jssln: via @ssorallen: This seems reasonable but changes the behavior of this except; it will no longer catch KeyboardInterrupts nor SystemExits.

ssorallen: I meant to say that that is a good thing. I can't think of why you would intentionally want to capture system exits and/or keyboard interrupts. I think we should merge this.

jssln: ah, got it. @aledista we will merge this!

aledista: :+1:

Issue #68, opened 2015-06-22
Collapse/Uncollapse all in file-tree


aewing:
The default behavior of the Atom file tree is to collapse/uncollapse a directory and all contained directories when the arrow icon is alt-clicked. Currently this is not happening for me in the default file tree on Ubuntu 14.04, Atom 0.205.0.

ssorallen:
This is now in master. alt-click on directories recursively expands or collapses them.

Commit 4f640638480bc1177863547cb453c99c257f6f6d added the functionality.

PR #127, opened 2015-07-05
bound the delete key to the file tree

### LINKED ISSUE #78: Bind the delete key to the file tree
Hitting `<delete>` on a file entry in the tree should trigger the file deletion dialog

edvinerikson:
Fixes #78

### DIFF
--- a/pkg/nuclide/file-tree/spec/FileTreeController-spec.js
+++ b/pkg/nuclide/file-tree/spec/FileTreeController-spec.js
@@ -15,6 +15,8 @@ var path = require('path');
 var rmdir = require('rimraf');
 var temp = require('temp').track();
 var timers = require('timers');
+var React = require('react-for-atom');
+var {TestUtils} = React.addons;
 
 var rootBasenames = [
   'dir1',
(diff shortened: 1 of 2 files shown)


ssorallen:
Thanks for the change. Would you mind adding a test case for this? Otherwise it looks good.

edvinerikson:
Does it look good now? 
I'm not very familiar with the code yet, so please point out if there is any better/smoother way to test the code.

pkg/nuclide/file-tree/spec/FileTreeController-spec.js
```diff
@@ -66,6 +66,21 @@ describe('FileTreeController', () => {
     fileTreeController.destroy();
   });
 
+  it('checks if core:backspace and core:delete is registered on .nuclide-file-tree', () => {
+    var timesCalled = 0;
+    // Find div element
+    var el = fileTreeController._panelController._hostEl.getElementsByClassName('nuclide-file-tree')[0];
+    // mock deleteSelection
+    fileTreeController.deleteSelection = () => {
```
ssorallen: Jasmine provides a concept for this called [spies](https://jasmine.github.io/1.3/introduction.html#section-Spies) that does roughly what you did but ensures the methods are restored after each spec. It would look something like this:

```
spyOn(fileTreeController, 'deleteSelection');

atom.commands.dispatch(el, 'core:backspace');
atom.commands.dispatch(el, 'core:delete');

expect(fileTreeController.deleteSelection.calls.length).toEqual(2);
```

pkg/nuclide/file-tree/spec/FileTreeController-spec.js
```diff
@@ -66,6 +66,21 @@ describe('FileTreeController', () => {
     fileTreeController.destroy();
   });
 
+  it('checks if core:backspace and core:delete is registered on .nuclide-file-tree', () => {
```
ssorallen: Thanks for adding `core:backspace` too. Can you split this into two specs? It might feel like duplication, but it'll be more instructive in the future if one of these commands break because a single, targeted spec will say exactly what failed.

edvinerikson: Thanks for the feedback! Will change it right away!

pkg/nuclide/file-tree/spec/FileTreeController-spec.js
```diff
@@ -66,6 +66,21 @@ describe('FileTreeController', () => {
...
+    var timesCalled = 0;
+    // Find div element
+    var el = fileTreeController._panelController._hostEl.getElementsByClassName('nuclide-file-tree')[0];
```
ssorallen: To keep to the public API of the FileTreeController, go through `getTreeComponent` and use React's [TestUtils](https://facebook.github.io/react/docs/test-utils.html#findrendereddomcomponentwithclass):

```javascript
var {TestUtils} = require('react-for-atom').addons;

...

var el = TestUtils. findRenderedDOMComponentWithClass(
  fileTreeController.getTreeComponent(),
  'nuclide-file-tree'
);
```

edvinerikson: I managed to get this work
```javascript
      var el = React.findDOMNode(TestUtils.findRenderedDOMComponentWithClass(
        fileTreeController._panelController.getChildComponent(),
        'nuclide-file-tree'
      ));
```
as `getTreeComponent` doesn't return `<div className="nuclide-file-tree">`

Is that ok?

ssorallen: Oh yeah, perfect. I forgot about the `findDOMNode` wrapper in that case. :100:

edvinerikson:
I think I got everything now. :sunglasses:

ssorallen:
Looks great, thanks for implementing the feedback. I'll pull this in for internal review and merge today or tomorrow. :+1:

edvinerikson:
Awesome! Thanks for all the feedback. :+1: 
I will continue to contribute with more features/bug fixing. Will probably start with those labeled `enhancement`.

ssorallen:
Thanks, we appreciate the work. I've been wanting https://github.com/facebook/nuclide/issues/68 implemented since working with larger file trees. That's another scoped feature that should be doable.

PR #131, opened 2015-07-06
added duplicate

### LINKED ISSUE #82: Add "Duplicate" to the file tree
This should work with both local and remote files.

edvinerikson:
Fixes #82 
Got keyboard shortcut on `d` (same as Atom's built-in file-tree)

### DIFF
--- a/pkg/nuclide/file-tree/keymaps/nuclide-file-tree.cson
+++ b/pkg/nuclide/file-tree/keymaps/nuclide-file-tree.cson
@@ -9,6 +9,7 @@
 
 '.nuclide-file-tree':
   'd': 'nuclide-file-tree:duplicate-selection'
+
 # TODO(jjiaa): Remove this after we figure out how to disable the tree-view package.
 'atom-workspace':
   'ctrl-alt-cmd-\\': 'tree-view:toggle'
--- a/pkg/nuclide/file-tree/keymaps/nuclide-file-tree.cson
+++ b/pkg/nuclide/file-tree/keymaps/nuclide-file-tree.cson
@@ -7,7 +7,8 @@
   'ctrl-\\': 'nuclide-file-tree:toggle'
   'ctrl-|': 'nuclide-file-tree:reveal-active-file'
 
-
+'.nuclide-file-tree':
+  'd': 'nuclide-file-tree:duplicate-selection'
 # TODO(jjiaa): Remove this after we figure out how to disable the tree-view package.
 'atom-workspace':
   'ctrl-alt-cmd-\\': 'tree-view:toggle'
--- a/pkg/nuclide/file-tree/lib/FileTreeController.js
+++ b/pkg/nuclide/file-tree/lib/FileTreeController.js
@@ -219,6 +219,7 @@ class FileTreeController {
           'nuclide-file-tree:add-folder': () => this.openAddFolderDialog(),
           'nuclide-file-tree:delete-selection': () => this.deleteSelection(),
           'nuclide-file-tree:rename-selection': () => this.openRenameDialog(),
+          'nuclide-file-tree:duplicate-selection': () => this.openDuplicateDialog(),
           'nuclide-file-tree:remove-project-folder-selection': () => this.removeRootFolderSelection(),
           'nuclide-file-tree:copy-full-path': () => this.copyFullPath(),
           'nuclide-file-tree:show-in-file-manager': () => this.showInFileManager(),
@@ -302,6 +303,13 @@ class FileTreeController {
         command: 'nuclide-file-tree:rename-selection',
         shouldDisplayForSelectedNodes(nodes) {
           return nodes.length === 1 && !nodes.some(node => node.isRoot());
+        },
+      },
+      {
+        label: 'Duplicate',
+        command: 'nuclide-file-tree:duplicate-selection',
+        shouldDisplayForSelectedNodes(nodes) {
+          return nodes.length === 1 && !nodes.some(node => node.getItem().isDirectory());
         },
       },
       {
@@ -745,6 +753,43 @@ class FileTreeController {
     };
     this._openDialog(<FileDialogComponent {...props} />);
   }
+
+  openDuplicateDialog(): void {
+    var selection = this._getSelectedEntryAndDirectoryAndRoot();
+    if (!selection) {
+      return;
+    }
+
+    if(selection.entry.isDirectory()) {
+      return;
+    }
+    var message = (
+      <div>
+        <div>Enter the new name for the file in the root:</div>
+        <div>{path.normalize(selection.root.getPath() + '/')}</div>
+      </div>
+    );
+
+    var {entry, root} = selection;
+
+    var FileDialogComponent = require('./FileDialogComponent');
+    var props = {
+      rootDirectory: root,
+      initialEntry: entry,
+      message,
+      onConfirm: async (rootDirectory, relativeFilePath) => {
+        var file = rootDirectory.getFile(relativeFilePath);
+        await file.create();
+        await entry.read().then(text => file.write(text));
+        this._reloadDirectory(entry.getParent());
+        atom.workspace.open(file.getPath());
+      },
+      onClose: () => this._closeDialog(),
+      shouldSelectBasename: true,
+    };
+    this._openDialog(<FileDialogComponent {...props} />);
+  }
+
 
   _openDialog(component: ReactElement): void {
     this._closeDialog();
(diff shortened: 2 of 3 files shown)


ssorallen:
Thanks for the PR. I'll take a look at this.

pkg/nuclide/file-tree/lib/FileTreeController.js
```diff
@@ -299,6 +300,13 @@ class FileTreeController {
         },
       },
       {
+        label: 'Duplicate',
+        command: 'nuclide-file-tree:duplicate-selection',
+        shouldDisplayForSelectedNodes(nodes) {
+          return nodes.length === 1 && !nodes.some(node => node.getItem().isDirectory());
```
ssorallen: Since the first assertion confirms the array is length 1, it can be accessed directly:

```js
return nodes.length === 1 && !nodes[0].getItem().isDirectory();
```

pkg/nuclide/file-tree/lib/FileTreeController.js
```diff
@@ -740,6 +748,43 @@ class FileTreeController {
...
+    var message = (
+      <div>
+        <div>Enter the new name for the file in the root:</div>
+        <div>{path.normalize(selection.root.getPath() + '/')}</div>
+      </div>
+    );
+
+    var {entry, root} = selection;
+
+    var FileDialogComponent = require('./FileDialogComponent');
```
ssorallen: Can you move this into `_openDialog`? It's a bit goofy that it was copied in the first place, we can do a little cleanup here. `_openDialog` can take `props` as an argument instead of a `ReactComponent`.

pkg/nuclide/file-tree/lib/FileTreeController.js
```diff
@@ -740,6 +748,43 @@ class FileTreeController {
     this._openDialog(<FileDialogComponent {...props} />);
   }
 
+  openDuplicateDialog(): void {
+    var selection = this._getSelectedEntryAndDirectoryAndRoot();
+    if (!selection) {
```
ssorallen: Can merge this and the next conditional:

```js
if (!selection || selection.entry.isDirectory())
```

pkg/nuclide/file-tree/keymaps/nuclide-file-tree.cson
```diff
@@ -7,7 +7,8 @@
   'ctrl-\\': 'nuclide-file-tree:toggle'
   'ctrl-|': 'nuclide-file-tree:reveal-active-file'
 
-
+'.nuclide-file-tree':
+  'd': 'nuclide-file-tree:duplicate-selection'
```
ssorallen: Linebreak after the end of objects.

edvinerikson:
Think I got everything now.

pkg/nuclide/file-tree/lib/FileTreeController.js
```diff
@@ -682,7 +690,6 @@ class FileTreeController {
       </div>
     );
 
-    var FileDialogComponent = require('./FileDialogComponent');
```
ssorallen: :+1: Thanks for this cleanup.

pkg/nuclide/file-tree/lib/FileTreeController.js
```diff
@@ -737,16 +743,50 @@ class FileTreeController {
...
+
+    var {entry, root} = selection;
+
+    var props = {
+      rootDirectory: root,
+      initialEntry: entry,
+      message,
+      onConfirm: async (rootDirectory, relativeFilePath) => {
+        var file = rootDirectory.getFile(relativeFilePath);
+        await file.create();
```
ssorallen: This needs to check the return value of [`create`](https://atom.io/docs/api/v1.0.0/File#instance-create) to see whether the file was successfully created or not. If it returns false, that means the file either couldn't be created or already existed.

ssorallen: If it couldn't be created, maybe create one more modal with the explanation that it couldn't be created with just an "OK" button? Any ideas there? We don't want to fail silently.

ssorallen:
This looks great, thank you. It needs to handle the case where the filename already exists, and then I think this should be good to go.

pkg/nuclide/file-tree/lib/FileTreeController.js
edvinerikson: Yeah, you are right, didn't think of that. Maybe a [`Notification`](https://atom.io/docs/api/v1.0.1/NotificationManager#instance-addError) can solve that?

ssorallen: :+1: A warning notification sounds good to me.

edvinerikson:
Does it look okay now? :sunglasses:

pkg/nuclide/file-tree/lib/FileTreeController.js
```diff
@@ -741,16 +747,62 @@ class FileTreeController {
...
+
+    var {entry, root} = selection;
+
+    var props = {
+      rootDirectory: root,
+      initialEntry: entry,
+      message,
+      onConfirm: (rootDirectory, relativeFilePath) => {
+        var file = rootDirectory.getFile(relativeFilePath);
+        file.create().then(async (createdSuccessfully) => {
```
ssorallen: This looks great, but let's stick to all async/await rather than switching between that and promises. I also reworded the detail message to remove the "We"; I find it eerie when a [computer](https://en.wikipedia.org/wiki/HAL_9000) responds like that :

```js
onConfirm: (rootDirectory, relativeFilePath) => {
  var file = rootDirectory.getFile(relativeFilePath);
  var createdSuccessfully = await file.create();
  if (createdSuccessfully) {
    await entry.read().then(text => file.write(text));
    this._reloadDirectory(entry.getParent());
    atom.workspace.open(file.getPath());
  } else {
    atom.notifications.addWarning(
      "Failed to duplicate file",
      {
        dismissable: true,
        detail: `There was a problem duplicating the file. Please check if the file "${file.getBaseName()}" already exists.`
      }
    );
  }
},
```

edvinerikson: Ah, yeah. Didn't know `var createdSuccessfully = await file.create();` was valid syntax.. It's changed now. :sunglasses:

ssorallen: Yeah `await` takes some practice to feel natural. This is a good rundown: https://github.com/yortus/asyncawait#8-await-in-depth-awaitable-expressions

pkg/nuclide/file-tree/spec/FileTreeController-spec.js
```diff
@@ -107,6 +107,23 @@ describe('FileTreeController', () => {
...
+        fileTreeController._panelController.getChildComponent(),
+        'nuclide-file-tree'
+      ));
+
+      // Mock method
+      spyOn(fileTreeController, 'openDuplicateDialog');
+      atom.commands.dispatch(el, 'nuclide-file-tree:duplicate-selection');
+      expect(fileTreeController.openDuplicateDialog.calls.length).toBe(1);
+    });
+    xit('fires an Notification if the file could not be created', () => {
```
ssorallen: Can you fill this in or delete it? The other dialog methods are also difficult to test, so deleting for now is fine if it's not easily testable.

ssorallen:
I'll pull this in to review internally. Thanks for all of this.

ssorallen:
I tried this locally and found one issue:

* Select multiple files in the file tree using Cmd/Ctrl + click. Press 'd'. The duplicate dialog pops up even though multiple files are selected. It should instead do nothing.

Also on improvement:

* Append "_copy" to the new filename so the name in the dialog is immediately a viable name for the new file. No need to check if the "_copy" file already exists, the duplicate error will catch that. This should make the filename unique *most* of the time though.

ssorallen:
Also:

* The new file opens correctly, but it is not selected in the file tree. It should be selected in the file tree after the duplicate is created.

pkg/nuclide/file-tree/lib/FileTreeController.js
edvinerikson: Ah, thanks for the link! :+1:

edvinerikson:
Think I got everything now. :) Will someone of you guys fix the commits (rewrite the history) or do I need to do it? (Never done it before)

ssorallen:
Can you rebase this onto master? Then this looks good to go.

edvinerikson:
Done! Thanks for all the feedback, I really need that to learn the codebase. :+1:

PR #132, opened 2015-07-06
Fix nuclide-start-server certificate validation

Keno:
The version check was failing due to certificate validation failures.
Fix this by:
- disabling host name validation
- passing the CA cert file to the client for validation

Also adds printing for unexpected SSL errors

### DIFF
--- a/pkg/nuclide/server/scripts/nuclide_server.py
+++ b/pkg/nuclide/server/scripts/nuclide_server.py
@@ -77,7 +77,7 @@ class NuclideServer(object):
             server_cert, server_key, ca = self.get_server_certificate_files()
             client_cert, client_key = self.get_client_certificate_files(ca)
             self._version = utils.http_get('localhost', self.port, method='POST', url='/server/version',
-                                           key_file=client_key, cert_file=client_cert)
+                                           key_file=client_key, cert_file=client_cert, ca_cert = ca)
         else:
             self._version = utils.http_get('localhost', self.port, method='POST', url='/server/version')
         return self._version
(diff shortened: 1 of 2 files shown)


Keno:
I see there was a similar pull request in #117, though this one does actually perform the validation. Also what is the minimum required python version for nuclide. It seems like we might need a version check here.

bolinfest:
@Keno just to save us the trouble of auditing: which APIs are you worried are not Python >= 2.6 compliant?

Keno:
the `context` argument to `HTTPSConnection` was introduced when cert validation was enabled by default, which happened in 2.7 I believe.

mtorromeo:
You are missing an import of `ssl`; other than this I'm in favor of merging this instead of my PR#117

Keno:
Updated with missing ssl import.

Keno:
Rebased and added a python version check.

Keno:
Looks like there's merge conflicts again. I'd be happy to rebase this, but please take a look so I can incorporate any review feedback while I'm at it.

cbosuna:
It looks good to me, but will test and get back to you Monday.

cbosuna:
Checked against several versions of python in both the 2.X series and the 3.X series and this patch works like a charm.  I think there is just one tiny merge conflict.  If you fix we can merge this pull request.

Keno:
Rebased.

Keno:
bump again

mtorromeo:
After 2 months, a fix this simple is not merged yet.
Even if one is willing to contribute you end up not even bothering anymore...

ssorallen:
@cbosuna Is there anything blocking this?

Keno:
Time for my monthly bump ;).

Keno:
Thanks, just in time ;)

PR #136, opened 2015-07-07
Added alt-click functionality to the file tree

### LINKED ISSUE #68: Collapse/Uncollapse all in file-tree

edvinerikson:
Fixes #68 
The test is a little bit flaky since the nodes used in the test doesn't actually represent those used in Atom.

### DIFF
--- a/pkg/nuclide/ui/tree/lib/TreeRootComponent.js
+++ b/pkg/nuclide/ui/tree/lib/TreeRootComponent.js
@@ -65,6 +65,18 @@ function toggleSetHas(
   return added;
 }
 
+function isPathIgnored(path: string): boolean {
+  var isIgnored = false;
+  atom.project.getRepositories().forEach(repo => {
+    if(!repo) {
+      return;
+    }
+    if(repo.isPathIgnored(path)) {
+      isIgnored = true;
+    }
+  });
+  return isIgnored;
+}
 var FIRST_SELECTED_DESCENDANT_REF: string = 'firstSelectedDescendant';
 
 /**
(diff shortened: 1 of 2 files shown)


pkg/nuclide/ui/tree/lib/TreeRootComponent.js
```diff
@@ -65,6 +65,18 @@ function toggleSetHas(
   return added;
 }
 
+function isPathIgnored(path: string): boolean {
```
ssorallen: This can be simplified with `some`:

```js
function isPathIgnored(path: string): boolean {
  return atom.project.getRepositories().some(repo => {
    return repo && repo.isPathIgnored(path);
  });
}
```

pkg/nuclide/ui/tree/lib/TreeRootComponent.js
```diff
@@ -178,6 +190,35 @@ var TreeRootComponent = React.createClass({
...
 
+  async _toggleChildNodesExpanded(parentNode: LazyTreeNode, forceExpanded: ?boolean): Promise {
+    if(isPathIgnored(parentNode.getItem().getPath())) {
+      return;
+    };
+    var childNodes = parentNode.getCachedChildren();
+
+    // fetch children if cache is empty
+    if(!childNodes) {
+      await parentNode.fetchChildren().then(nodes => childNodes = nodes);
```
ssorallen: Avoid single-line arrow functions if the function body is an assignment because single line arrows implicitly return the value of the expression.

```js
await parentNode.fetchChildren().then(nodes => {
  childNodes = nodes;
});
```

pkg/nuclide/ui/tree/lib/TreeRootComponent.js
```diff
@@ -205,7 +246,14 @@ var TreeRootComponent = React.createClass({
   },
 
   _onClickNodeArrow(event: SyntheticEvent, node: LazyTreeNode): void {
-    this._toggleNodeExpanded(node);
+    // if alt-clicked the arrow, expand child directories
+    if(event.altKey) {
+      // show/hide child nodes when clicked
+      this._toggleChildNodesExpanded(node, !this._isNodeExpanded(node))
```
ssorallen: The ordering here will make expanding a large, remote directory spin for a long time before giving feedback in the UI because this waits to expand the clicked directory until all of its children have expanded. Instead, the clicked directory should expand immediately and then its children should be recursively expanded.

```js
_onClickNodeArrow(...) {
  this._toggleNodeExpanded(node);
  if (event.altKey) {
    this._toggleChildNodesExpanded(node, this._isNodeExpanded(node));
  }
}

File: pkg/nuclide/ui/tree/lib/TreeRootComponent.js [hunk 4]
```diff
@@ -178,6 +190,35 @@ var TreeRootComponent = React.createClass({
     this.setState({expandedKeys});
   },
 
+  async _toggleChildNodesExpanded(parentNode: LazyTreeNode, forceExpanded: ?boolean): Promise {
```
ssorallen: Check out `expandNodeKey` and `collapseNodeKey`; I think the functionality there can be reused to reduce the amount of new code that needs to be rewritten here. Expanding all children might be better as options to those methods to limit new methods on this class too.

edvinerikson: Not sure if I fully understands what you mean here. 
Do you want me to remove the `_toggleChildNodesExpanded` method and change `expandNodeKey` and `collapseNodeKey`to take a argument (boolean) which will expand/collapse the child nodes?

ssorallen: That seems like a better approach because those methods already implement expanding/collapsing. They could be recursively expand children of the node rather than implement that in a separate function.

ssorallen:
Closing for now based on discussion. This should reuse more of the code that already exists in `TreeRootComponent` because the functionality is largely the same.

PR #142, opened 2015-07-13
fix(ui-tabs): remove unused object-assign

teppeis:
Remove unused `Object.assign` and `object-assign` package.


ssorallen:
Nice find. Thanks for this.

PR #171, opened 2015-08-18
[hack] Minor cleanup

steelbrain:
Linter doesn't trigger lint until the file has a valid path. So we can just ignore the check here.

### DIFF
--- a/pkg/nuclide/hack/lib/HackLinter.js
+++ b/pkg/nuclide/hack/lib/HackLinter.js
@@ -18,11 +18,6 @@ module.exports = {
   scope: 'file',
   lintOnFly: true,
   async lint(textEditor: TextEditor): Promise<Array<Object>> {
-    var file = textEditor.getBuffer().file;
-    if (!file) {
-      return [];
-    }
-
     var diagnostics = await findDiagnostics(textEditor);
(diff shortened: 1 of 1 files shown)


mostafaeweda:
I guess we can do that now since we moved away from the broken `linter` package.
Thanks for contribution, will definitely test and take that in.

mostafaeweda:
Our hack linter, based on a hack worker can only process and lint files that has a valid path.
Hence, what this won't really fix it, but what would fix the usage for you is to save the hack file as `.php` or `.hh` file and every hack language feature should work fine with you.

steelbrain:
@mostafaeweda I am not sure if I had explained accurately what I was trying to. This is a check that avoids linting files that have not been saved yet (`textEditor.getPath()` returns empty).
I have already moved this check to the core `linter`, You can just skip it in the providers.

mostafaeweda:
@steelbrain we ae moving away from the `linter` anyways for the sake of the new `diagnostics` package. This is just a safety check that ensures whatever linter or diagnostics versions we use, the hack linter specifically won't lint `null` or `undefined` path files/editors.

steelbrain:
Understandable.

PR #213, opened 2015-10-06
Fix linter-eslint

LegNeato:
Either the latest version of linter-eslint or linter doesn't
create a Range object correctly. Work around the issue.

### DIFF
--- a/pkg/nuclide/diagnostics/store/lib/LinterAdapter.js
+++ b/pkg/nuclide/diagnostics/store/lib/LinterAdapter.js
@@ -50,6 +50,8 @@ export type LinterProvider = {
   lint: (textEditor: TextEditor) => Promise<Array<LinterMessage>>;
 };
 
+import {Range} from 'text-buffer';
+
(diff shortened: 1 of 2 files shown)


LegNeato:
Without this Nuclide throws an error that it can't read `row` on `range.start` and I don't get `eslint` messages from `linter-eslint` in Nuclide's UI.

With this:

screen shot 2015-10-05 at 8 53 08 pm

LegNeato:
I wanted to fix it the correct way but don't have time today. Also, curious on your thoughts on the `"*"` dep being better for this case. (don't know why PR comments seem to have disappeared? Because I force pushed?)

nmote:
Somebody internal just put together a diff that fixes this, so I'm going to close this pull request. Thanks for taking the time to put it up, though!

Regarding the `*` dep -- firstly, 'text-buffer' actually isn't needed, since we can just `require('atom')` to get `Range`. Anyway, we want to control our dependencies well and since we don't use `npm shrinkwrap` across the board yet we can at least get part of the way by hardcoding version numbers. That way we at least get known versions of our direct dependencies whenever we `npm install`.

PR #222, opened 2015-10-25
List Hyperclick providers

AsaAyers:
I don't see anything else on apm that seems to use hyperclick.

### DIFF
--- a/pkg/nuclide/hyperclick/README.md
+++ b/pkg/nuclide/hyperclick/README.md
@@ -6,10 +6,15 @@ Hyperclick is triggered by two events:
 - `<alt>` or `<alt-mousemove>` underlines clickable text under the mouse.
 - `<alt-mousedown>` performs the action associated with the clickable text.
 - `<cmd-alt-enter>` performs the action on the text under the cursor.
-- 
+-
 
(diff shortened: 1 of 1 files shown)


bolinfest:
The following packages use hyperclick:

https://atom.io/packages/nuclide-buck-files
https://atom.io/packages/nuclide-clang-atom
https://atom.io/packages/nuclide-flow
https://atom.io/packages/nuclide-hack
https://atom.io/packages/nuclide-ocaml

In particular, `nuclide-flow` gets its information about JavaScript from Flow. Where does the information from `js-hyperclick` for determining location definitions come from?

AsaAyers:
`js-hyperclick` uses babel to parse the files. It collects the set of scopes, references to external modules (`import` and `require()`), and exports (`module.exports = `, `export`, and `export default`). To avoid constant parsing it holds an internal cache that is invalidated on `onDidStopChanging` and is only recreated the first time hyperclick requests a suggestion on something that might be an identifier when the grammer is JavaScript.

As I see the bot mentioned, I updated the PR to include the items you listed. But as I don't know anything about them I didn't add descriptions.

PR #231, opened 2015-10-30
alt click doesn't work.

yomybaby:
on OSx, alt click doesn't work now.
But on your readme.md, alt + click is supported.

cmd(meta) click is conflict with multi cursor click.

https://github.com/facebooknuclideapm/hyperclick/blob/master/README.md

### DIFF
--- a/pkg/nuclide/hyperclick/lib/HyperclickForTextEditor.js
+++ b/pkg/nuclide/hyperclick/lib/HyperclickForTextEditor.js
@@ -10,7 +10,7 @@
  */
 
 import type Hyperclick from './Hyperclick';
-import {trackTiming, startTracking} from 'nuclide-analytics';a
+import {trackTiming, startTracking} from 'nuclide-analytics';
 import type {TimingTracker} from 'nuclide-analytics';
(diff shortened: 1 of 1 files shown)


ssorallen:
Because the meta key is the accelerator commonly used in other IDEs, Nuclide intentionally chose to use it as well. Ideally Atom would default the multi-cursor accelerator to the alt key or make it configurable so [hyperclick](https://atom.io/packages/hyperclick) could use the meta key.

Going to close this for now since the functionality is intentional.

bolinfest:
I thought we [historically, at least] exposed this via a setting?

nmote:
Yeah, I think that exposing this via a setting would be reasonable, but I agree that the current behavior is desirable. The documentation should be updated to include the OS X behavior, though.

There is an Atom issue (https://github.com/atom/atom/issues/7595) for deactivating multiple cursors. If we had our way, multiple cursors would move to a different keybinding and no longer conflict.

nmote:
(We don't currently expose it as a setting but I would happily review a PR that did)

Issue #254, opened 2015-11-26
nuclide-hack stops autocomplete from within the current file from working


bakert:
Steps to reproduce:

0. Install and enable nuclide-hack package.

1. Open a new file.

2. Add:

    <?hh

    class MyTest {
    }

3. Save this file as MyTest.php.

4. Type:

    $myTest = new MyT

5. Engage autocomplete either via Ctrl-Space or Cmd-Shift-P and choose off the menu.

Expected:

* Autocomplete fire automatically in step 4 and show MyTest as an option.
* Autocomplete will show MyTest as an option when fired manually.

Actual:

* Autocomplete does not believe it has a suggestion.

Tested on 1.3.0-beta6 and 1.0.18. Same problem if you use `<?php` instead of `<?hh`. This works fine if nuclide-hack package is disabled.

Some other interesting notes:

* If MyTest is defined in another file that is currently open it will be a suggestion to autocomplete. Even if the new file has not been saved and contains no PHP-ish symbols.
* If you perform the exact same test but save the file as MyTest.js it will work.

bakert:
I don't have HHVM installed locally (only remotely) and thus I don't have hh_client on my path. That's suggested by the README. I wonder if that would help? Doesn't seem related. Console shows no errors, autocomplete just stops working.

bakert:
Removing this section of nuclide-hack's config.json solves the problem:

    "autocomplete.provider": {
      "versions": {
        "2.0.0": "createAutocompleteProvider"
      }
    },

PR #264, opened 2015-12-08
Don't ignore lower pri suggestions in nuclide-hack autocomplete.

bakert:
This fixes https://github.com/facebook/nuclide/issues/254

I don't know why this option was originally set to true. Fixing this may introduce other issues. I'm happy to dig deeper if you have any pointers.


nmote:
I'm inclined to approve this. We don't exclude lower priority results for our Flow autocomplete provider and personally I think it's fine. @mostafaeweda, What was the original rationale for excluding lower priority results? Will something go wrong if we land this?

mostafaeweda:
yea, let's take this in - I've just verified that nothing wrong would come from that.

PR #265, opened 2015-12-08
Adding extra globals

fabiomcosta:
including some browser builtins and some JS globals


pkg/nuclide/format-js-base/lib/constants/builtIns.js
```diff
@@ -17,24 +17,51 @@ module.exports = new Set([
   'Boolean',
   'Date',
   'Error',
+  'Function',
+  'Infinity',
   'JSON',
   'Math',
   'Number',
   'Object',
+  'Promise',
```
kyldvs: I guess this is fine now to not require in www, right @DmitrySoshnikov ?

pkg/nuclide/format-js-base/lib/constants/builtIns.js
```diff
@@ -17,24 +17,51 @@ module.exports = new Set([
...
+  'performance',
+  'prompt',
+  'screen',
+  'sessionStorage',
+  'setInterval',
+  'setTimeout',
   'window',
+  
+  'Option',
+  'Image',
```
kyldvs: I think this one will conflict with `Image.react` which we do want to require. Can we remove this?

pkg/nuclide/format-js-base/lib/constants/builtIns.js
```diff
@@ -17,24 +17,51 @@ module.exports = new Set([
...
+  'open',
+  'performance',
+  'prompt',
+  'screen',
+  'sessionStorage',
+  'setInterval',
+  'setTimeout',
   'window',
+  
+  'Option',
```
kyldvs: What is `Option`?

pkg/nuclide/format-js-base/lib/constants/builtIns.js
DmitrySoshnikov: Yep, should be global now (auto-imported module).

pkg/nuclide/format-js-base/lib/constants/builtIns.js
DmitrySoshnikov: https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option

pkg/nuclide/format-js-base/lib/constants/builtIns.js
fabiomcosta: :+1:

kyldvs:
LGTM, someone else will need to merge it in

fabiomcosta:
I forgot `navigator` :|

PR #276, opened 2015-12-15
git clone address

joanna-chen:
the previous address directed to a permission error


ssorallen:
Thanks for the change. The permission error is likely because you don't have an SSH key uploaded to GitHub, but even so the 'https' URLs are the recommended clone URLs in GitHub's documentation: https://help.github.com/articles/which-remote-url-should-i-use/ We will switch to the https one as well.

PR #283, opened 2015-12-17
Update instructions on developing from source

ssorallen:


zertosh:
You don't really need Python

ssorallen:
@zertosh I can remove Python. I updated the Atom version to at least 1.0 since I don't think we should claim support for such an old version. We might want to adopt an evergreen support policy with Atom: most recent patch release of the current and previous minor versions. That's for another change and pull request though.

zertosh:
:+1:

PR #285, opened 2015-12-17
Add instructions for uninstalling prev packages

ssorallen:
Version 0.0.35 was the last version to be many packages. Add
instructions for uninstalling the former set of packages.


ssorallen:
Gist embedding is supported in GitHub's flavor of Jekyll+plugins. Here's the rendered version:

screen shot 2015-12-17 at 2 36 29 pm

PR #294, opened 2015-12-29
Nuclide git support

viatsko:
These commits add git support to nuclide's file-tree sub-package and add nuclide-blame-provider-git package.


viatsko:
Oups, I'll put it in a separate branch and will reopen pull request

Issue #274, opened 2015-12-15
Failed to activate the nuclide-hg-repository package


prasannamuley:
[Enter steps to reproduce below:]

1. Followed Nuclide under a single package doc
2. After starting Atom first time, have this issue.

**Atom Version**: 1.3.1
**System**: Microsoft Windows 10 Home
**Thrown From**: [nuclide-hg-repository](https://github.com/facebook/nuclide) package, v0.0.0

### Stack Trace

Failed to activate the nuclide-hg-repository package

```
At Invalid regular expression: /\/: \ at end of pattern

SyntaxError: Invalid regular expression: /\/: \ at end of pattern
  at new RegExp (native)
  at Object.<anonymous> (E:\Projects\nuclide\pkg\nuclide\hg-repository-client\lib\utils.js:18:22)
  at Module._compile (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\native-compile-cache.js:92:30)
  at Object.defineProperty.value [as .js] (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\compile-cache.js:208:21)
  at Module.load (module.js:355:32)
  at Function.Module._load (module.js:310:12)
  at Module.require (module.js:365:17)
  at require (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\native-compile-cache.js:41:21)
  at Object.<anonymous> (E:\Projects\nuclide\pkg\nuclide\hg-repository-client\lib\HgRepositoryClient.js:29:14)
  at Module._compile (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\native-compile-cache.js:92:30)
  at Object.defineProperty.value [as .js] (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\compile-cache.js:208:21)
  at Module.load (module.js:355:32)
  at Function.Module._load (module.js:310:12)
  at Module.require (module.js:365:17)
  at require (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\native-compile-cache.js:41:21)
  at Object.<anonymous> (E:\Projects\nuclide\pkg\nuclide\hg-repository-client\lib\main.js:11:27)
  at Module._compile (C:\Users\prasa\AppData\Local\atom\app-1.3.1\resources\app.asar\src\native-compile-cache.js:92:30)
[...]

PR #297, opened 2015-12-30
Fix for #274

viatsko:


zertosh:
I had a different fix for this and few other windows problems. It was something along the lines of `modifiedPath.split(path.sep)`. But I'll accept this for now :+1:

PR #298, opened 2015-12-30
Added nuclide-blame-provider-git

viatsko:


mostafaeweda:
This would need to be refactored to use the service framework to be able to use for remote projects as well.

mostafaeweda:
look at `HgService.js` and `FileSystemService.js` and usages in nuclide's code base as a reference to use the service framework.

viatsko:
Hi! Thanks for the feedback, will do. Same for async support for git-utils! We want git remote projects internally in company too

mostafaeweda:
Great to hear, would be good as well to sync with @nathansobo & @maxbrunsfeld about Atom's async GitRepository implementation, hopefully we can do it once or at least reuse parts of it.

nathansobo:
@joshaber is actually the most knowledgable about the new async repo.

viatsko:
Yes, we already discussed that with him. I'll work on a patch for nuclide on top of async branch this week.

mostafaeweda:
@viatsko would be even cooler to have git compatibility with the 'Diff Viewer' through implementing the called methods on `_repository` here: https://github.com/facebook/nuclide/blob/master/pkg/nuclide/diff-view/lib/RepositoryStack.js

viatsko:
:+1: I already have some stuff working, still a lot of stuff to fix

Also another 2 cents: we need to have repository links pattern config option, what do you think? (people using gitweb, github, gitlab etc and I'm not sure that everybody's able to create their own *./fb/config*)

mostafaeweda:
Sure, we can create a default `config` that's merged with `fb/config` if that exists.

viatsko:
:+1: would be nice

viatsko:
Another question, this one is related to nuclide-server - it looks like in both implementations we should compile binaries - for git-utils (libgit2) or nodegit, in Atom's case we can use bundled libraries (and actually npm install will cause binary libs to fail otherwise), but for nuclide-server should we include those libraries installation or maybe we can go with git shell commands execution for the first iteration? That could be easier and I'll finish those today then. Otherwise, I'll need suggestions on libgit2/nodegit + nuclide-server if shell exec is not good for initial iteration :)

mostafaeweda:
I understand libgit2, git-utils & nodegit are synchronous and we want to avoid that trap (I think the original GitRepository implementation uses that), hence, I believe the git shell commands implementation is our best option, we do that for mercurial as well, see the implementation.

mostafaeweda:
Actually, nodegit is not synchronous (it's now used in Atom's new async git implementation: https://github.com/atom/atom/blob/master/src/git-repository-async.js & see how to get the blame info from it here: https://github.com/nodegit/nodegit/blob/master/test/tests/blame.js#L31)

viatsko:
I see, but we need to make it working on server side also.

The problem with Atom's Git now is that the default git module is overriding custom one because of race condition (you'll always get default provider on editor's load and your custom git provider 2+ project opening). I'll try to look into it in the nearest weeks to finally make git usage possible in Nuclide.

mostafaeweda:
@viatsko we had a discussion about that recently and I'd love if you could hack around that (maybe even replacing the original instance somehow after nuclide loads).

aiboy:
Any update on that?

nmote:
Feel free to reopen if more work is done here

PR #299, opened 2015-12-30
Added git support to tree-view

viatsko:
In one commit, cleaner implementation, will work on further improvements


mostafaeweda:
I'd like to defer this until Atom's `GitRepository` implementation is all async API and serviceable, so, we can use for local and remote development.

Atom have promised that's an in-progress project to move away from the sync nature of the git repo implementation.

PR #308, opened 2016-01-11
Added ability to skip `./module_cycle_check` in scripts/dev/test

viatsko:
Otherwise it will fail on ./fb/setup and ./fb/config files hard.


viatsko:
Hi, any other suggestions? Having this flag makes tests working without pkg/fb/ directory

johnislarry:
Thanks, looks good, we'll merge this in.  Also 

https://github.com/facebook/nuclide/commit/f95c0ffdd0b565fdd89485ebc518a1d50cb9ac98

fixes the original issue

PR #335, opened 2016-01-27
Fix ocamlmerlin config option

chenglou:
Most likely the leftover of the old nuclide setup? Plain
`'nuclide-ocaml.pathToMerlin'` isn't a thing unless you've explicitly
installed that package, which is deprecated afaik. This fixes the config
path.

There are other related places:

- pkg/nuclide/buck/base/lib/BuckProject.js:106
- pkg/nuclide/commons/lib/systemInfo.js:36

(New to Atom & Nuclide, not sure what the proper `nuclide-features-config ` should have been. But this is a quick fix).


nmote:
Thanks! This is fine. I wish I could justify spending more time on this package. I'd love to have really good OCaml support but we just don't have enough people writing OCaml to justify working on this over other things. If you're keen on spending some time here there is a lot of low hanging fruit.

PR #348, opened 2016-02-02
Correct URL for Flow & Hack pages in docs

danharper:


JoelMarcey:
@danharper Thanks for finding this. 

Can you actually make the change from `master` instead of `gh-pages`. There is a `docs` directory in `master` that gets sync'd with `gh-pages`. In essence, `gh-pages` is read-only for us. I should clarify that.

danharper:
Sure no problem, #351.

PR #350, opened 2016-02-03
Expose underlying Merlin Process.

jordwalke:
This hooks is very useful for prototyping on cutting edge Merlin services which are not yet made official.

### DIFF
--- a/pkg/nuclide/ocaml-base/lib/LocalMerlinService.js
+++ b/pkg/nuclide/ocaml-base/lib/LocalMerlinService.js
@@ -48,3 +48,15 @@ export async function complete(
   const instance = await getInstance(path);
   return instance ? instance.complete(path, line, col, prefix) : null;
 }
+
+/**
+ * Low-level API into merlin service useful for debugging and for prototyping
+ * on top of bleeding edge Merlin branches.
+ */
+export async function runSingleCommand(
+  path: NuclideUri,
+  command: mixed
+): Promise<any> {
+  const instance = await getInstance(path);
+  return instance ? instance.runSingleCommand(command) : null;
+}


nmote:
I'm pretty sure this won't work if you are using Merlin over a remote connection. The service framework will try to serialize the Merlin Process and will probably fail. I'll merge this provided that:

- You add a comment noting that this will not work remotely and is only for local prototyping. Committed code should not use this function.
- You test to make sure that the existing features still work remotely. I'm pretty sure that the service framework won't choke on this definition but it has bitten me before.

Lastly, thanks for hacking on this! I really want our OCaml tooling to be better but haven't found the time to improve it. I'm hoping this is a sign that there are more contributions to come :smile:

jordwalke:
I'll be hopefully adding more and more features for OCaml development into Nuclide. I'll be targeting a new unreleased version of Merlin though which is why I must tap into the lower level protocol.

jordwalke:
How would I do this such that it works remotely as well?

ssorallen:
@jordwalke It won't work remotely because Nuclide's service framework doesn't know how to serialize a Merlin process instance. 'LocalMerlinService.js' acts as the service definition for the MerlinService, and calls to functions exported from the module that are correctly Flow annotated with primitive types can be serialized. Serialization is handled in the [proxy-generator](https://github.com/facebook/nuclide/blob/c4414ad7f3303b76d9d89d4a016fbb7705a07a3f/pkg/nuclide/service-parser/lib/proxy-generator.js).

You could wrap calls to the Merlin process with annotated, exported functions in LocalMerlinService so the calls can be serialized across the service framework properly.

jordwalke:
I'm still learning about Nuclide. Why can't local services just be remote services that happen to point to localhost. Seems like that architecture would make developing for Nuclide much easier.

ssorallen:
@jordwalke It works like it sounds like you're describing. When the LocalMerlinService process is requested locally, it delegates to a local instance and does no serialization; you should get a reference to the Merlin process object if you run this locally.

If you want to run this on the server and make calls to it locally do you need to worry about serialization. The service framework is designed to make the existence of serialization transparent to the developer, but it relies on well-annotated services that use a defined set of types that the service framework knows how to serialize.

```javascript
const {getServiceByNuclideUri} = require('../../client');

// Because '/some/path' is a local path, `localMerlinService` will delegate to a local instance
// of MerlinService and perform no serialization.
const localMerlinService = await getServiceByNuclideUri('MerlinService', '/some/path');

// Because 'nuclide://some/path' is remote, `remoteMerlinService` will delegate to the service
// framework and perform serialization of calls according to the proxy generated by parsing
// the type annotations in 'LocalMerlinService.js'
const remoteMerlinService =
  await getServiceByNuclideUri('MerlinService', 'nuclide://some/path');

// Calling methods on them looks exactly the same though.
const localLocation = await localMerlinService.locate('/another/path', 15, 25, 'foo');
const remoteLocation = await remoteMerlinService.locate('/another/path', 15, 25, 'foo');
```

jordwalke:
I wish there was a way to develop without having to worry about testing in two modes. Why not just have a way to go through the serialization path even when developing locally? This lowers the barrier to contributing.

ssorallen:
> I wish there was a way to develop without having to worry about testing in two modes.

That's a fair point, and in other pull requests from open source contributors we have commented saying, "but remote won't work". Maybe in development mode (starting Atom with `atom --dev`) Nuclide could always create proxies so you'll get the serialization errors during development.

jordwalke:
Yes, that would go a long ways. There are people (such as myself the first time I looked into contributing), who try contributing and then before submitting the pull request, they back out because there's this completely other mode which they don't understand.

ssorallen:
@jordwalke :+1: Looking into this now and if it would be possible. Thanks for your feedback.

jordwalke:
I updated this PR so that the arguments are future-proof serializable in the event that some serializable automation is added. It doesn't look like it matters because this is not serialized IIRC.

jordwalke:
Does this look good to go?

ssorallen:
@jordwalke Yeah this looks good. I'll import this internally.

PR #389, opened 2016-03-07
Allow configurable flags to ocamlmerlin

jordwalke:
There are some useful flags in `ocamlmerlin`. Let's expose the ability to customize them, at least for debugging purposes.

### DIFF
--- a/pkg/nuclide/ocaml-base/lib/MerlinProcess.js
+++ b/pkg/nuclide/ocaml-base/lib/MerlinProcess.js
@@ -235,6 +235,7 @@ export async function getInstance(file: NuclideUri): Promise<?MerlinProcess> {
   }
 
   const merlinPath = getPathToMerlin();
+  const flags = getMerlinFlags();
 
   if (!await isInstalled(merlinPath)) {
     return null;
@@ -247,7 +248,7 @@ export async function getInstance(file: NuclideUri): Promise<?MerlinProcess> {
   };
 
   logger.info('Spawning new ocamlmerlin process');
-  const process = await safeSpawn(merlinPath, [], options);
+  const process = await safeSpawn(merlinPath, flags, options);
   merlinProcessInstance = new MerlinProcess(process);
 
   if (dotMerlinPath) {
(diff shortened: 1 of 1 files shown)


pkg/nuclide/ocaml-base/lib/MerlinProcess.js
```diff
@@ -269,6 +270,18 @@ function getPathToMerlin(): string {
...
 
+/**
+ * @return The set of arguments to pass to ocamlmerlin.
+ */
+function getMerlinFlags(): Array<String> {
+  const configVal = global.atom
+    && global.atom.config.get('nuclide.nuclide-ocaml.merlinFlags');
+  // To split while stripping out any leading/trailing space, we match on all
+  // *non*-whitespace.
+  const configItems = configVal && configVal.match(/\S+/g);
```
zertosh: What do you think about `require('shell-quote').parse`? https://github.com/substack/node-shell-quote

jordwalke: I can swap it out if you like - especially if there's precedence for using this utility. I don't care either way, because there are so few arguments that are typically passed to `ocamlmerlin` at startup time.

jordwalke: I looked into that library and its probably overkill for merlin's use case (since in practice there's only one/two command line flag that you'd actually use and the rest should appear project scoped to a `.merlin` library).
I'm happy to add it if you feel strongly.

pkg/nuclide/ocaml-base/lib/MerlinProcess.js
```diff
@@ -269,6 +270,18 @@ function getPathToMerlin(): string {
     && global.atom.config.get('nuclide.nuclide-ocaml.pathToMerlin') || 'ocamlmerlin';
 }
 
+/**
+ * @return The set of arguments to pass to ocamlmerlin.
+ */
+function getMerlinFlags(): Array<String> {
+  const configVal = global.atom
+    && global.atom.config.get('nuclide.nuclide-ocaml.merlinFlags');
```
nmote: This won't work at all remotely. I'm fine with adding this, so long as you don't write code in the future that depends on certain flags having been passed.

nmote:
Overall looks fine to me. I don't have an opinion about shell quoting so I will leave the rest to @zertosh.

pkg/nuclide/ocaml-base/lib/MerlinProcess.js
zertosh: We have `shell-quote` already and the Clang package uses it for this exact purpose. But I don't feel that strongly about it. :+1:

PR #391, opened 2016-03-07
Increase hit target for header links

vjeux:
We want the entire block to be clickable, not just the text itself

screen shot 2016-03-07 at 2 49 59 pm


JoelMarcey:
Thanks @vjeux!

I accidentally merged this directly from GitHub. I need to bring it internally. So I had to revert. But I will get it in there. :)

PR #399, opened 2016-03-11
[ocaml-base] put opambin into PATH before spawning merlin

yunxing:
Sometimes people don't run `opam config env` before starting Atom. Therefore opam's binary directory (something like `/Users/yunxing/.opam/4.02.3/bin`) may not be included in the search path and Nuclide cannot find merlin to run.

This patch forces opam to be part of PATH before running merlin. 

@jordwalke

### DIFF
--- a/pkg/nuclide-ocaml-base/lib/MerlinProcess.js
+++ b/pkg/nuclide-ocaml-base/lib/MerlinProcess.js
@@ -145,6 +145,7 @@ export class MerlinProcess {
     });
   }
 
+
   async enclosingType(
     file: NuclideUri,
     line: number,
@@ -229,6 +230,22 @@ export class MerlinProcess {
   }
 }
 
+/**
+ * @return A promise of an augmented environment variables with
+ *   PATH that contains opam binary directory
+ */
+async function getEnvWithOpamBin(): Promise<String> {
+  const ret = await checkOutput('opam', ['config', 'var', 'bin']);
+  if (ret.exitCode === 0) {
+    const env = {...process.env};
+    env.PATH = `${env.PATH}${path.delimiter}${ret.stdout.trim()}`;
+    return env;
+  } else {
+    logger.error('Cannot get opam bin');
+    return process.env;
+  }
+}
+
 let merlinProcessInstance: ?MerlinProcess;
 
 export async function getInstance(file: NuclideUri): Promise<?MerlinProcess> {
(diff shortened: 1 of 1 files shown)


pkg/nuclide/ocaml-base/lib/MerlinProcess.js
```diff
@@ -284,12 +304,22 @@ function getMerlinFlags(): Array<string> {
 
 let isInstalledCache: ?boolean = null;
 async function isInstalled(merlinPath: string): Promise<boolean> {
+  opamBin = await getOpamBin();
```
jordwalke: Should probably be a little utility function.

pkg/nuclide/ocaml-base/lib/MerlinProcess.js
```diff
@@ -243,8 +257,14 @@ export async function getInstance(file: NuclideUri): Promise<?MerlinProcess> {
 
   const dotMerlinPath = await findNearestFile('.merlin', file);
 
+  opamBin = await getOpamBin();
+  env = process.env.PATH;
+  if (opamBin) {
+    env = env + ":" + opamBin
```
jordwalke: Is there a cross platform way to join paths? I genuinely do not know.

ssorallen: They can be joined like strings here, but the platform-specific delimiter should be used from Node's [`path.delimiter`](https://nodejs.org/api/path.html#path_path_delimiter):

```js
import path from 'path';

...

if (opamBin) {
  env = `${env}${path.delimiter}${opamBin}`;
}
```

ssorallen:
Can you try linting your code? It looks like you're missing `const` or `let` on your variable declarations, which is something ESLint will catch.

```
$ cd /path/to/nuclide
$ npm install
$ npm run lint-all
```

pkg/nuclide/ocaml-base/lib/MerlinProcess.js
yunxing: Good point!

yunxing:
@ssorallen @jordwalke Linted and fixed as suggested.

yunxing:
@ssorallen @jordwalke Does this look good?

ssorallen:
Looks good to me. Can you rebase onto latest master?

@jordwalke, look good to you?

jordwalke:
Yeah awesome thanks.

nmote:
@jordwalke @ssorallen are either of you planning on driving this through?

nmote:
@yunxing if you rebase one more time I will drive this through. Sorry that we dropped the ball on this PR.

nmote:
I'm still happy to make sure this gets in if you update. Feel free to reopen or create a new PR.

jordwalke:
I think we've got an even better solution coming up! :D

PR #418, opened 2016-03-18
first commit of the redone file tree filter

AlexJuarez:


mostafaeweda:
Could you add some description explaining what are you trying to do?

AlexJuarez:
This is an update to prefix interaction that shows the prefix as you type, highlights matches and isn't case sensitive.

ssorallen:
Hey thanks for the pull request. I haven't yet had a chance to try this, but I will in the next few days.

PR #419, opened 2016-03-19
[code-format] Switch between formatCode and formatEntireFile

chenglou:
This addresses the issue from
https://github.com/facebook/nuclide/commit/73b8c7346aa419bc4f05216ab578aeffb3f617a2#commitcomment-15965354
where we check for the presence of either formatCode or formatEntireFile
without taking into consideration whether there's a selected region.

The future logic for this should probably change, but this current
behaviour is _mostly_ backward-compatible.

Test plan:

Try it on a formatter that provides both options.

Small table (true = present):
Old:

| formatCode | formatEntireFile | result |
| ---------- | ---------------- | ------ |
| true | true | formatcode |
| true | false | formatcode |
| false | true | formatEntireFile |
| false | false | error |

New:

| formatCode | formatEntireFile | selectionEmpty | result |
| ---------- | ---------------- | -------------- | ------ |
| true | true | true | formatEntireFile (small behaviour change) |
| true | true | false | formatCode |
| true | false | true | formatCode |
| true | false | false | formatCode |
| false | true | x | formatEntireFile |
| false | false | x | error |

### DIFF
--- a/pkg/nuclide-code-format/lib/CodeFormatManager.js
+++ b/pkg/nuclide-code-format/lib/CodeFormatManager.js
@@ -49,7 +49,8 @@ class CodeFormatManager {
     const selectionRange = editor.getSelectedBufferRange();
     const {start: selectionStart, end: selectionEnd} = selectionRange;
     let formatRange = null;
-    if (selectionStart.isEqual(selectionEnd)) {
+    const selectionRangeEmpty = selectionRange.isEmpty();
+    if (selectionRangeEmpty) {
       // If no selection is done, then, the whole file is wanted to be formatted.
       formatRange = buffer.getRange();
     } else {
@@ -62,7 +63,8 @@ class CodeFormatManager {
     }
 
     const provider = matchingProviders[0];
-    if (provider.formatCode != null) {
+    if (provider.formatCode != null &&
+      (!selectionRangeEmpty || provider.formatEntireFile == null)) {
       const codeReplacement = await provider.formatCode(editor, formatRange);
       // TODO(most): save cursor location.
       editor.setTextInBufferRange(formatRange, codeReplacement);
--- a/pkg/nuclide-external-interfaces/1.0/atom.js
+++ b/pkg/nuclide-external-interfaces/1.0/atom.js
@@ -374,6 +374,7 @@ declare class atom$Range {
   constructor(pointA: RangeConstructorArg, pointB: RangeConstructorArg): void;
   start: atom$Point;
   end: atom$Point;
+  isEmpty(): boolean;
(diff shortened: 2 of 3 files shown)


pkg/nuclide-external-interfaces/1.0/atom.js
```diff
@@ -374,6 +374,7 @@ declare class atom$Range {
   constructor(pointA: RangeConstructorArg, pointB: RangeConstructorArg): void;
   start: atom$Point;
   end: atom$Point;
+  isEmpty(): boolean;
```
chenglou: This API exists since the earliest documented version of Atom (1.3.2)

pkg/nuclide-code-format/lib/CodeFormatManager.js
```diff
@@ -62,7 +63,11 @@ class CodeFormatManager {
     }
 
     const provider = matchingProviders[0];
-    if (provider.formatCode != null) {
+    // If formatCode exists, use it if selection isn't empty, or if it's empty
+    // but formatEntireFile doesn't exist.
+    if (provider.formatCode != null &&
+      (!selectionRangeEmpty ||
+        (selectionRangeEmpty && provider.formatEntireFile == null))) {
```
hansonw: Simplify the boolean expression! :P

`!A || A && B === !A || B`

hansonw:
Thanks @chenglou! Mostly looks good, just:
1. simplify the logic (see inline comment) nuclide-code-format/lib/types.js#L13
2. change the docblock here to document this behaviour: https://github.com/chenglou/nuclide/blob/format-opts/pkg/nuclide-code-format/lib/types.js#L13

pkg/nuclide-code-format/lib/CodeFormatManager.js
chenglou: But that'd actually be less readable and wouldn't shortcut (it currently does), lol

hansonw: what do you mean? the second "selectionRangeEmpty" check is redundant, so it actually avoids an extra check - to clarify, what I mean is

```
if (provider.formatCode != null && (!selectionRangeEmpty || provider.formatEntireFile == null))) {
```

IMO it's just as readable - the explanation would be "use me if the range is non-empty, or we're forced to because `formatEntireFile` isn't defined."

chenglou: Oh oops, misread what you wrote. Fixed!

chenglou:
Done!

hansonw:
:+1:

chenglou:
No I didn't

PR #435, opened 2016-03-29
Fix local projects on Windows.

### LINKED ISSUE #375: Nuclide file tree removes current file tree and shows nothing
### Issue and Steps to Reproduce
Problem
Tree view breaks and can never get it to show files/folders.
Empty space with add project folder is shown

Solution: 
Went to nuclide settings and disable the following 2 settings:  

1. Enable the "nuclide-file-tree" feature
Can not find '' inside the View tab to even display it
Removes your current file tree

2. Enable the "nuclide-side-bar" feature
This creates a big blank area next to the file tree (if first is disabled, otherwise shows nothing), which is where Im sure the nuclide file tree would have been shown 

### Expected Behavior
Tree view should be shown

### Actual Behavior
Tree view does not show, and does not allow you to add any project folders

### Versions
* Atom: 0.119
* Nuclide: 0.119.0
* Client OS: Windows 10

### Edit: Pictures added

Both Disabled
image

Both Enabled
image
image

Tree disabled, but sidebar enabled
image

File Tree enabled, sidebar disabled
image

Daniel15:
Nuclide should use [path.isAbsolute](https://nodejs.org/api/path.html#path_path_isabsolute_path) to determine if a local file path is absolute rather than checking if `path.indexOf(0) === separator`, as that approach doesn't work on Windows.

Closes #375

### DIFF
--- a/pkg/nuclide-file-tree/lib/FileTreeContextMenu.js
+++ b/pkg/nuclide-file-tree/lib/FileTreeContextMenu.js
@@ -13,7 +13,7 @@ import {CompositeDisposable} from 'atom';
 import {EVENT_HANDLER_SELECTOR} from './FileTreeConstants';
 import FileTreeStore from './FileTreeStore';
 
-import {isFullyQualifiedLocalPath} from './FileTreeHelpers';
+import path from 'path';
 
 type MenuItemSingle = {
   label: string;
@@ -196,7 +196,7 @@ class FileTreeContextMenu {
     const node = this._store.getSingleSelectedNode();
     return (
       node != null &&
-      isFullyQualifiedLocalPath(node.nodePath) &&
+      path.isAbsolute(node.nodePath) &&
       process.platform === platform
     );
   }
--- a/pkg/nuclide-file-tree/lib/FileTreeHelpers.js
+++ b/pkg/nuclide-file-tree/lib/FileTreeHelpers.js
@@ -139,17 +139,13 @@ function getDisplayTitle(key: string): ?string {
 function isValidDirectory(directory: Directory): boolean {
   return (
     !isLocalEntry((directory: any)) ||
-    isFullyQualifiedLocalPath(directory.getPath())
+    pathModule.isAbsolute(directory.getPath())
   );
 }
 
 function isLocalEntry(entry: Entry): boolean {
   // TODO: implement `RemoteDirectory.isRemoteDirectory()`
   return !('getLocalPath' in entry);
-}
-
-function isFullyQualifiedLocalPath(path: string): boolean {
-  return path.charAt(0) === pathModule.sep;
 }
 
 function isContextClick(event: SyntheticMouseEvent): boolean {
@@ -176,7 +172,6 @@ module.exports = {
   getDisplayTitle,
   isValidDirectory,
   isLocalEntry,
-  isFullyQualifiedLocalPath,
   isContextClick,
   buildHashKey,
 };
--- a/pkg/nuclide-file-tree/spec/FileTreeHelpers-spec.js
+++ b/pkg/nuclide-file-tree/spec/FileTreeHelpers-spec.js
@@ -75,17 +75,16 @@ describe('FileTreeHelpers', () => {
   });
 
   describe('on Windows', () => {
-    let originalPathSep;
+    let originalPathModule;
 
     beforeEach(() => {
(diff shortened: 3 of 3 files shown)


ssorallen:
Excellent, thank you for this. I was in the middle of downloading VirtualBox and a Windows VM to finally fix this because my hunch was it was something small and path-related. I'll import this internally and get it landed.

bolinfest:
Will `path.isAbsolute()` properly exclude remote paths?

pkg/nuclide-file-tree/lib/FileTreeContextMenu.js
```diff
@@ -13,7 +13,7 @@ import {CompositeDisposable} from 'atom';
 import {EVENT_HANDLER_SELECTOR} from './FileTreeConstants';
 import FileTreeStore from './FileTreeStore';
 
-import {isFullyQualifiedLocalPath} from './FileTreeHelpers';
+import {isAbsolute} from 'path';
```
ssorallen: The pattern in Nuclide is to import the path module as 'path' and call functions on it.

```
import path from 'path';

...

path.isAbsolute(node.nodePath) &&
```

Daniel15:
@bolinfest - It appears to:

```
> var path = require('path');
undefined
> path.isAbsolute('c:\\src\\')
true
> path.isAbsolute('c:\\src\\foo.txt')
true
> path.isAbsolute('nuclide://foo')
false
> path.isAbsolute('nuclide://foo/bar')
false
>
```

pkg/nuclide-file-tree/lib/FileTreeContextMenu.js
Daniel15: Thanks, fixed.

ssorallen:
This breaks a spec I added when I ensured remote development worked on Windows: https://github.com/facebook/nuclide/blob/7537bb8a09b1ae13db07a7dfb8ca71bd92aeb3c1/pkg/nuclide-file-tree/spec/FileTreeHelpers-spec.js#L126 You can run the specs yourself by going into the 'pkg/nuclide-file-tree' directory and running `apm test`. You need to run `apm install` from the root first if you haven't already.

I'm unsure what effect that break will have. Do remote directories still work after this change?

Daniel15:
@ssorallen - Hmm, I think remote directories still worked, I'll have to double-check tonight though.

I think the issue is that the Node.js `path` module has different behaviour based on the operating system it's running on. Consequently, this means the behaviour of `isValidDirectory` is dependent on OS, and the spec is not taking that into account. It's checking if `\a\b\c` is a valid directory, which untrue on every OS. It should be checking `/a/b/c` on non-Windows systems, and `c:\a\b\c` on Windows.

I think the ideal fix is to have a test that properly handles multiple OSes:
```
if (on Linux) {
  runLinuxTests();
} else (on Windows) {
  runWindowsTests();
} else (on Mac OS X) {
  runMacOSXTests();
}
```

And then ensure the build runs on every operating system.

Alternatively, I wonder if there's a way to force Node.js to use the Win32 implementation of path.js.

Daniel15:
[path.js](https://github.com/nodejs/node/blob/master/lib/path.js) has this at the end:

```
posix.win32 = win32.win32 = win32;
posix.posix = win32.posix = posix;


if (process.platform === 'win32')
  module.exports = win32;
else
  module.exports = posix;
```

So `require('path')` gets you the implementation for the current platform, and you can do `require('path').win32` or `require('path').posix` to get a particular implementation.

I wonder if there's a way to swap out the `require('path')` so it explicitly requires the win32 version for unit testing (that is, what would be returned by `require('path').win32`). This is where a proper dependency injection framework would come in handy.

Any suggestions for what the test should do to handle this?

Daniel15:
@ssorallen - Just tried one approach to fix the test, let me know what you think. Rather than just overriding `path.sep` like it did before, I override the entire `path` module. This means the Windows versions of `path.isAbsolute` and friends are used. If you squint, it almost looks like dependency injection. :stuck_out_tongue: 

```
Finished in 10.975 seconds
69 tests, 135 assertions, 0 failures, 0 skipped
```

Also confirming that remote projects work. See screenshot below - "nuclide" is a local project while "live" is a remote one:

ssorallen:
@Daniel15 I dig the "dependency injection." I'll pull in your changes, make sure everything works on non-Windows, and get this landed. Thanks for fixing the test.

Issue #437, opened 2016-03-29
Outline View should support Function Expressions


DaleJefferson:
### Issue and Steps to Reproduce
![screen shot 2016-03-29 at 23 38 51](https://cloud.githubusercontent.com/assets/375005/14126236/6d1ab290-f607-11e5-9154-e9f912845d18.png)

```
/* @flow */

function functionDeclaration() {
}

var functionExpression = function () {
};
```

### Expected Behaviour
I would expect both types to show up in the Outline View

### Actual Behaviour
Only the function declarations is displayed

### Versions
* Atom: 1.6.0
* Nuclide: 0.127.0
* Client OS: Mac OS

### Additional Details
* Installed packages (`apm ls --installed`):

├── atom-typescript@8.4.0
├── autocomplete-modules@1.4.1
├── editorconfig@1.2.4
├── git-plus@5.13.0
├── haskell-grammar@0.4.0
├── highlight-selected@0.11.2
├── language-babel@2.16.0
├── language-ini@1.15.0
├── language-ocaml@1.1.2
├── language-thrift@1.0.2
├── linter@1.11.4
├── linter-eslint@7.1.3
├── nuclide@0.127.0
├── pigments@0.26.0
├── sort-lines@0.14.0
├── terminal-plus@0.14.5
└── tool-bar@0.4.0

nmote:
Yep, there are a bunch of different things that I'm planning to support with outline view in the near future but I wanted to get an initial working version off the ground ASAP. I'll probably get to this in the next two weeks.

nmote:
exports.foo =..., module.exports = { ... }, and maybe top-level constants are some of the others.

DaleJefferson:
@nmote It looks like a really promising feature, it seems to do a really good job with ES6 classes. I eagerly await wider language support, thanks for adding this feature.

PR #441, opened 2016-03-31
Show class fields in flow outline.

pvolok:
This new outline feature is awesome. But it does not show fields for classes. I think it's a common pattern to define some methods as fields like this:

```js
class Component {
  state = {};

  render() {
    return <button onClick={this.handleClick} />;
  }

  handleClick = (event) => {
  };
}
```

It would be nice if such methods would be shown as well. This PR draft makes above class to appear like this:

```
class Component
  state=
  render()
  handleClick=(event)
```

If you are cool with this change, I'll add tests.


nmote:
See also #437 for a few other things I want to support, if you are feeling inspired to keep working on this.

I'm not totally sure how I feel about displaying the equals sign but I'm willing to land it and see how it works in practice.

Update the tests and I'll merge.

pvolok:
I added tests for class fields. Unfortunately I'm short on time this week and won't yet be able to address the issue you linked. Maybe later :)

nmote:
@facebook-github-bot import

nmote:
I made two minor tweaks -- removed trailing whitespace to pass lint, and changed the test file to `@noflow` since it now has Flow errors. Those were both needed to pass the pre-commit tests and I didn't want to have another back-and-forth before landing.

Thanks for putting this together! Totally understand re: being busy. I just wanted to make you aware of the other plans in case you were interested.

PR #445, opened 2016-04-02
Add troubleshooting information

markmiro:
These are some troubleshooting steps that I go through. I'm not confident that this is the best process. The last step did something but killing node and python was just a guess. I had this come up once and the fix worked, so not sure.

Hope this helps people. I rely on Nuclide for work.


JoelMarcey:
@markmiro Sorry it has taken so long to get to this. 

We have a troubleshooting doc that this might fall better into.

https://nuclide.io/docs/help/troubleshooting/

The file is at: https://github.com/facebook/nuclide/blob/master/docs/_docs/help/troubleshooting.md

JoelMarcey:
Closing in favor of #558

PR #490, opened 2016-05-11
Keymap to create new file with tree view path

marduke182:
I always try not to use the mouse when i'm coding and right now if I need create a new file I must use mouse right click `new -> file` option.

The attempt of this pull request is allow file creation only with the keyboard for example `ctrl-0` and `cmd+n`


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @zertosh to be a potential reviewer.

zertosh:
cc: @shushz

shushz:
I don't think its a good thing to change the Cmd-n behavior. You can always do it in your keymaps file if you find it more convenient, but I think that turning it on by default with Nuclide might lead to an unnecessary confusion.

Notice that core atom also doesn't do it.

marduke182:
They use key **a** in the original tree view to add a new file. But nuclide use this key to search pattern.

For more information original tree view https://github.com/atom/tree-view/blob/master/keymaps/tree-view.cson#L74

PR #512, opened 2016-05-24
Fix issue: empty string -> no valid JSON error

frogcjn:
This change will fix meaningless error:
```shell
[ERROR] nuclide - Invalid JSON result from flow status --json /Users/frogcjn/Files/Projects/Node.js-Projects/shoutbox-babel/src/index.js. JSON:
```
when value is an empty string.

### DIFF
--- a/pkg/nuclide-flow-base/lib/FlowRoot.js
+++ b/pkg/nuclide-flow-base/lib/FlowRoot.js
@@ -331,7 +331,7 @@ export class FlowRoot {
 
 function parseJSON(args: Array<any>, value: string): any {
   try {
-    return JSON.parse(value);
+    return value ? JSON.parse(value) : {};;
   } catch (e) {
     logger.error(`Invalid JSON result from flow ${args.join(' ')}. JSON:\n'${value}'.`);
     throw e;
--- a/pkg/nuclide-flow-base/lib/FlowRoot.js
+++ b/pkg/nuclide-flow-base/lib/FlowRoot.js
@@ -331,7 +331,7 @@ export class FlowRoot {
 
 function parseJSON(args: Array<any>, value: string): any {
   try {
-    return value ? JSON.parse(value) : {};;
+    return value ? JSON.parse(value) : {};
   } catch (e) {
     logger.error(`Invalid JSON result from flow ${args.join(' ')}. JSON:\n'${value}'.`);
     throw e;


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @nmote, @zertosh and @matthewwithanm to be potential reviewers.

facebook-github-bot-6:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-8:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

pkg/nuclide-flow-base/lib/FlowRoot.js
```diff
@@ -331,7 +331,7 @@ export class FlowRoot {
 
 function parseJSON(args: Array<any>, value: string): any {
   try {
-    return JSON.parse(value);
+    return value ? JSON.parse(value) : {};;
```
ssorallen: Double semicolon here.

nmote:
I agree that we should get rid of this error, but I don't think this is the right way to go about it. This is just a band-aid on top of the actual problem, which is Flow returning invalid JSON (I think due to a timeout? I haven't investigated but maybe increasing the number of retries or something would be a good fix). We shouldn't just silence the error, we should fix the root cause.

frogcjn:
I think the problem is that if there is no error, then the parameter `value` is empty, then JSON.parse('') will report an error.

nmote:
I understand that -- but the issue is that we end up with an empty string here in the first place. That shouldn't happen and it really does indicate a problem. It's unfortunate that this happens so often that it leads to console spam, but every time it happens it means there was a request that goes unfulfilled. We should fix the root cause rather than just silencing this message.

facebook-github-bot-7:
@frogcjn updated the pull request.

frogcjn:
@nmote
Its source is `result = await this._process.execFlow(args, options, file, /* waitForServer */ true);`
with the error of
"Fatal error: exception End_of_file\n"

frogcjn:
Its should be https://github.com/facebook/flow/issues/937

nmote:
If you'd like to increase timeouts or retries I'll take another look but I don't believe ignoring this error is the right way to solve this.

PR #527, opened 2016-06-01
Tell flow to ignore missing facebook modules

freebroccolo:
This PR tells `flow` not to report errors for the various places where `fb` modules are required but may not be available. I'm assuming these are modules for internal use at facebook since they are not installed with the other dependencies and, as far as I could tell, not mentioned in the documentation.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -21,6 +21,8 @@
 suppress_comment=.*\\$FlowFixMe.*
 suppress_comment=.*\\$FlowIssue.*
 suppress_comment=.*\\$FlowIgnore.*
+# uncommenting the next line will silence flow errors about missing 'fb' modules
+# suppress_comment=.*\\$FlowFB.*
 
 esproposal.class_static_fields=enable
 esproposal.decorators=ignore
--- a/pkg/commons-node/passesGK.js
+++ b/pkg/commons-node/passesGK.js
@@ -22,6 +22,7 @@ export default async function passesGK(
   // Only do the expensive require once.
   if (gatekeeper === undefined) {
     try {
+      // $FlowFB
       gatekeeper = require('../fb-gatekeeper').gatekeeper;
     } catch (e) {
       gatekeeper = null;
--- a/pkg/hyperclick/lib/main.js
+++ b/pkg/hyperclick/lib/main.js
@@ -23,6 +23,7 @@ export function activate() {
   // with the Hyperclick "confirm-cursor" command.
   // TODO(hansonw): Remove when symbols-view has a proper API.
   try {
+    // $FlowFB
     const {overrideGoToDeclaration} = require('./fb/overrideGoToDeclaration');
     overrideGoToDeclaration();
   } catch (e) {
--- a/pkg/nuclide-analytics/lib/track.js
+++ b/pkg/nuclide-analytics/lib/track.js
@@ -17,6 +17,7 @@ export type TrackEvent = {
 // This extra module enables adding spies during testing.
 export let track;
 try {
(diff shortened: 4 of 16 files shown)


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @hansonw, @zertosh and @peterhal to be potential reviewers.

nmote:
Unfortunately this will break our internal copy: the files do exist, and Flow will complain of an unused suppression. Since we run Flow pre-commit we can't take this as-is.

One thing I've thought about is pulling all of the module names out into a separate string on the line above, and then calling `require` with that string, e.g.:

```
const gatekeeperModule = '../fb-gatekeeper';
// $FlowIgnore
gatekeeper = require(gatekeeperModule).gatekeeper;
```

This way Flow will complain that "The parameter passed to require() must be a literal string" either way, so we can safely FlowIgnore it whether or not the module exists. This is kind of gross but I don't think it's any worse than what we have.

CC @zertosh, @peterhal -- does this seem reasonable?

nmote:
@zertosh points out that this will make us lose the type info for our internal modules so my idea is a non-starter.

nmote:
So here's an even worse idea, but I'm not sure I can get it past code review: put a blatant type error on the same line as the require so there is always something to ignore.

```
// $FlowIgnore
gatekeeper = require('../fb-gatekeeper').gatekeeper; ('': number)
```

However it's been a goal to eliminate these requires inside try/catches altogether so maybe we should just wait until we get to that...

freebroccolo:
What about using a different comment pattern for these cases? Maybe something like this:

```javascript
// $FlowFBRequire
gatekeeper = require('../fb-gatekeeper').gatekeeper;
```

If you used a special comment pattern, you could either swap out the .flowconfig file or edit it with `sed` or something like that for the internal case where you don't want flow to ignore it.

nmote:
I don't think we have the infrastructure to have a different .flowconfig synced to this repo, nor, I think, do we want to. Keeping it up-to-date and properly merged would be irritating. Anyway, I believe that right now all we can do is exclude files from getting synced, but other than that everything here is exactly the same as our internal copy.

freebroccolo:
Having to maintain a separate file internally would be a hassle, I agree. I'm thinking maybe you would have to go that far though. Is there some way to conditionally set certain directives in the `.flowconfig` file from the `flow` command or anything like that?

Even if there isn't, if a special comment pattern like `$FlowFBRequire` were used here instead of `$FlowIgnore`, you could add a line to the `.flowconfig` file like this:

```editorconfig
[options]

suppress_comment=.*\\$FlowFixMe.*
suppress_comment=.*\\$FlowIssue.*
suppress_comment=.*\\$FlowIgnore.*
# suppress_comment=.*\\$FlowFBRequire.*
```

That way external users would have an easy way to ignore these errors without causing any trouble in the default case for internal users. You could then either just mention that this line should be uncommented for external users in the documentation or you could provide an option to one of the scripts which would do it automatically.

freebroccolo:
I've implemented this last suggestion. I think this approach shouldn't cause any problems for internal users and it's easy enough for me to just uncomment the one line in `.flowconfig`.

facebook-github-bot-0:
@freebroccolo updated the pull request.

nmote:
This seems pretty reasonable to me.

@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/142107596202805/int_phab) to review.

PR #544, opened 2016-06-13
reorder comparison operators to maximize ligature coverage

hillbrad:
In the original, the = match from keyword.operator.assignment.php will fire before !=, ===, etc. can be matched by keyword.operator.comparison.php.  This splits the comparison operator into two styled spans, which breaks ligature replacement in code-oriented fonts.  This diff reorders more specific matches for keyword.operator.comparison.php ahead of the less specific ones for keyword.operator.assignment.php.

### DIFF
--- a/pkg/nuclide-language-hack/grammars/hack.cson
+++ b/pkg/nuclide-language-hack/grammars/hack.cson
@@ -1181,6 +1181,10 @@
         'name': 'keyword.operator.error-control.php'
       }
       {
+        'match': '(!==|!=|===|==|<=|>=|<>|<|>)'
+        'name': 'keyword.operator.comparison.php'
+      }
+      {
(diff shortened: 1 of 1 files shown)


nmote:
Seems pretty reasonable to me. @peterhal, @most, can you take a look?

mostafaeweda:
the tests aren't passing with that change - I had to split to:
```js
{
  'match': '(!==|!=|===|==)'
  'name': 'keyword.operator.comparison.php'
}
{
  'match': '=|\\+=|\\-=|\\*=|/=|%=|&=|\\|=|\\^=|<<=|>>='
  'name': 'keyword.operator.assignment.php'
}
{
  'match': '(<=|>=|<>|<|>)'
  'name': 'keyword.operator.comparison.php'
}
```
to fix the handling of `<<=` and `>>=` while matching `!=` and `===` as `keyword.operator.comparison.php`

PR #538, opened 2016-06-08
Correct menu options path to Panes in docs

simon-johansson:
Minor fix... At least the Panes options are under `View | ...` on OSX

PR #549, opened 2016-06-16
Remove Panes section from docs

simon-johansson:
Same fix as in #538 but now rebased onto master instead


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @JoelMarcey and @matthewjohnston4 to be potential reviewers.

PR #558, opened 2016-06-24
Update troubleshooting.md

markmiro:
These are some troubleshooting steps that I go through. I'm not confident that this is the best process. The last step did something but killing node and python was just a guess. I had this come up once and the fix worked, so not sure.


facebook-github-bot-0:
By analyzing the blame information on this pull request, we identified @JoelMarcey, @matthewjohnston4 and @zertosh to be potential reviewers.

markmiro:
See https://github.com/facebook/nuclide/pull/445

PR #612, opened 2016-07-15
Hyperclick uses ctrl-key instead of alt-key on Windows

markvandenbrink:
It seems the Hyperclick package uses the ctrl-key instead of the described alt-key on Windows. Using the ctrl-key is a problem in Atom, because it is used for making multiple text/cursor selections. This PR changes the ctrl-key to alt-key.

### DIFF
--- a/pkg/hyperclick/lib/HyperclickForTextEditor.js
+++ b/pkg/hyperclick/lib/HyperclickForTextEditor.js
@@ -332,7 +332,7 @@ export default class HyperclickForTextEditor {
    */
   _isHyperclickEvent(event: SyntheticKeyboardEvent | MouseEvent): boolean {
     // If the user is pressing either the meta/ctrl key or the alt key.
-    return process.platform === 'darwin' ? event.metaKey : event.ctrlKey;
+    return process.platform === 'darwin' ? event.metaKey : event.altKey;
   }
 
   _doneLoading(): void {


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @nmote, @mostafaeweda and @zertosh to be potential reviewers.

facebook-github-bot-6:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

johnislarry:
Where does it say that Hyperclick on Windows should be alt+click?  @mostafaeweda has done some work in the past to try and improve the competition between multiple cursors and Hyperclick.  The current behavior is:

When there are Hyperclick results ready, it's assumed you want to use Hyperclick instead of using the multi-cursor feature -- so, the event is prevented from propagating up triggering the multi-cursor feature.

However, when the Hyperclick has no results, e.g. clicking at the end of the line, or on non-symbols we delegate to the multi-cursor feature.

We don't want to change away from ctrl/cmd+click for Hyperclick because it's a keybinding that a lot of editors choose.  One solution is to make it clearer when Hyperclick has given up or is otherwise not doing "Hyperclick" work, and has delegated to the multi-cursor functionality.

markvandenbrink:
Ok, I understand and I know other editors use the familiar ctrl+click. Don't want to interfere with previous work. But the README confused me, as it states:

> Hyperclick is triggered by two events:
> - `<alt>` or `<alt-mousemove>` underlines clickable text under the mouse.
> - `<alt-mousedown>` performs the action associated with the clickable text.
> - `<cmd-alt-enter>` performs the action on the text under the cursor.

So I thought the alt-key was the preferred key. I often use the multi-selection/cursor feature in Atom to select items which do have Hyperclick results. I need to disable the Hyperclick package to do that.

It probably would be better to improve Hyperclick by disabling it in the following cases:
- There is a text selection active;
- When there are 2 or more text cursors active, because the editor is in multi-selection mode.

dsifford:
I agree with @markvandenbrink -- ctrl is used as the default key on Linux as well and it totally throws off multi-cursor. 

I'd toss my vote in the "allow users to choose the default key" bucket. Specifically, the mod key for Linux users would be perfect.

Thanks for your work on this nevertheless. Very, very useful. 👍

piuccio:
> allow users to choose the default key

👍 

The doc says `<alt>` but on my mac it works with `<cmd>` which is annoying when you're trying to make multiple selections.

PR #639, opened 2016-07-28
Restrict max workers to something sane

billschaller:
When the number of CPUs on a machine is quite large, flow spawns enough processes to saturate IO and starve everything else on the machine. This change makes it so that on machines with a very large number of cores, flow plays more nicely with other things (arc, hh, etc) happening concurrently. This is especially impactful when you do a rebase that touches a lot of files, and try to run a lot of unit tests or something shortly after.

See https://www.desmos.com/calculator/or5ppybxia for a graph of current versus new algos for max workers.

### DIFF
--- a/pkg/nuclide-flow-base/lib/FlowProcess.js
+++ b/pkg/nuclide-flow-base/lib/FlowProcess.js
@@ -310,7 +310,7 @@ export class FlowProcess {
   }
 
   _getMaxWorkers(): number {
-    return Math.max(os.cpus().length - 2, 1);
+    return Math.ceil(os.cpus().length / 4);
   }
 
   /**


billschaller:
t12428823

facebook-github-bot-4:
By analyzing the blame information on this pull request, we identified @nmote, @zertosh and @hansonw to be potential reviewers.

billschaller:
Also: I did some measurement - on a >32-core machine

max_workers | seconds elapsed for flow check
---------------|-----------------------------------
4 | 365
8 | 197
16 | 131
32 | 119

nmote:
I want to be very careful about making Flow slower. It's an important part of JS development for people who use it, and while it's rechecking it can't provide any IDE features. Here are a few other ideas that I would like to investigate before just chopping more cores off blindly:

* Create a config option with a sane default so that people can specify their own preferred number of cores.
* You say this is a problem with big machines in particular. So maybe make the heuristic take that into account and use a smaller percentage of cores on a bigger machine.
* Run Flow with `ionice` and/or `nice` when available.
* You cite rebases as a specific pain point. Maybe Flow could wait to recheck until after a rebase is complete (we watch for rebases in Nuclide and delay work until they are over, so there is precedent). Obviously this one would require changes to Flow.

billschaller:
Yes, running flow with nice would be nice. I did not test that specifically, but right now it's running at the default nice value. 

Even with nice though, "Use all but 2 cores" is pretty extreme. If you graph the dropoff in runtime gains, you see that amdahl's law kicks in pretty quick on pretty much any feasible system, even with enterprise SSDs. Even something more conservative like x/2 or x^(.66) would be a better default than what it's doing now.

nmote:
Maybe on 32 core machines, but if you have 4 or 8 cores you aren't going to be drowning in IO nearly as much. Which is why one of the suggestions I made is just a more complex heuristic. In your example I think using 16/32 cores would be reasonable, but I don't want to use 4/8 on my laptop.

And ionice ought to solve the IO problem altogether without slowing down the CPU-bound parts.

mkonicek:
> I did some measurement - on a >32-core machine

@billschaller Is this when changing just one file and saving (quite common case), or analyzing a large codebase? Does Flow make use of more than one worker when just one file changed?

I'm curious if 4 max workers would be sufficient for most cases when people edit a few files, save and want to see Flow errors highlighted as quickly as possible.

nmote:
> Does Flow make use of more than one worker when just one file changed?

Yes. The number of files changed is not a good predictor of the number of workers Flow needs. It has to recheck every file that is dependent on the changed file(s). I've seen that number approach 1000 in large codebases. There are several phases but the local inference/typechecking phase is trivially parallelizable I believe. If you are checking 1000 files it's going to be slow but you definitely want to throw as much computation at it as you can spare in that case.

I recommend getting in touch with the Flow team if you need a more detailed answer. But generally-speaking we want all the workers we can get.

nmote:
@billschaller A while back I made the change to run Flow with nice/ionice. You said you would follow up and I haven't heard anything about it from you, so I assume that alleviated your pain? With that in mind I'm going to close this.

PR #642, opened 2016-08-01
Add outline provider for OCaml.

pvolok:
It's hard to navigate in Flow's gigantic files :)

### DIFF
--- a/pkg/nuclide-ocaml-rpc/lib/MerlinService.js
+++ b/pkg/nuclide-ocaml-rpc/lib/MerlinService.js
@@ -33,6 +33,14 @@ export type MerlinError = {
   type: 'type' | 'parser' | 'env' | 'warning' | 'unknown',
 };
 
+export type MerlinOutline = {
+  start: MerlinPosition,
+  end: MerlinPosition,
+  kind: string,
+  name: string,
+  children: Array<MerlinOutline>,
+};
+
 export async function pushDotMerlinPath(path: NuclideUri): Promise<?any> {
   const instance = await getInstance(path);
   return instance ? instance.pushDotMerlinPath(path) : null;
@@ -87,6 +95,11 @@ export async function errors(
 ): Promise<?Array<MerlinError>> {
   const instance = await getInstance(path);
   return instance ? instance.errors() : null;
+}
+
+export async function outline(path: NuclideUri) {
+  const instance = await getInstance(path);
+  return instance ? instance.outline(path) : null;
 }
 
 /**
--- a/pkg/nuclide-ocaml-rpc/lib/MerlinService.js
+++ b/pkg/nuclide-ocaml-rpc/lib/MerlinService.js
@@ -97,7 +97,9 @@ export async function errors(
   return instance ? instance.errors() : null;
 }
 
-export async function outline(path: NuclideUri) {
+export async function outline(
+  path: NuclideUri,
+): Promise<?Array<MerlinOutline>> {
   const instance = await getInstance(path);
   return instance ? instance.outline(path) : null;
 }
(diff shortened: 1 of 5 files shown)


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @peterhal, @zertosh and @hansonw to be potential reviewers.

facebook-github-bot-3:
@pvolok updated the pull request.

pkg/nuclide-ocaml-rpc/lib/MerlinService.js
```diff
@@ -89,6 +97,11 @@ export async function errors(
   return instance ? instance.errors() : null;
 }
 
+export async function outline(path: NuclideUri) {
```
nmote: This needs a return type, since this is part of the RPC boundary. If you run the tests they should fail because of this.

pkg/nuclide-ocaml/lib/OutlineProvider.js
```diff
@@ -0,0 +1,90 @@
...
+      break;
+    case 'Signature':
+      kind = 'sig';
+      nameToken = className(data.name);
+      break;
+    case 'Type':
+      nameToken = type(data.name);
+      break;
+  }
+  kind = kind || data.kind.toLowerCase();
```
nmote: We prefer checking for null/undefined explicitly rather than relying on truthiness. For example if `kind` is `''` this will have confusing results. It's slightly more verbose but

```
if (kind != null) {
  kind = data.kind.toLowerCase();
}
```

captures the intention more exactly.

pkg/nuclide-ocaml/lib/OutlineProvider.js
```diff
@@ -0,0 +1,90 @@
...
+  nameToken = nameToken || plain(data.name);
```
nmote: Same here

pkg/nuclide-ocaml/lib/OutlineProvider.js
```diff
@@ -0,0 +1,90 @@
...
+  if (!instance) {
+    return null;
+  }
+  await instance.pushNewBuffer(path, editor.getText());
+  const result = await instance.outline(path);
+  if (!Array.isArray(result)) {
+    return null;
+  }
+  return {
+    outlineTrees: result.reverse().map(convertMerlinOutline),
```
nmote: It's surprising that this needs to be reversed. I'm assuming that's the way merlin reports it? Could the `reverse` be pushed further down the stack to keep the higher-level APIs cleaner?

pkg/nuclide-ocaml/lib/main.js
```diff
@@ -41,6 +42,16 @@ export function provideLinter(): LinterProvider {
   return MerlinLinterProvider;
 }
 
+export function provideOutlines(): OutlineProvider {
+  const {getOutline} = require('./OutlineProvider');
```
nmote: You can just `import {getOutline} from './OutlineProvider';` at the top of the file. We have a babel transformation in place that will not actually `require` the file until `getOutline` is referenced so there is no performance benefit to lazy-requiring explicitly.

nmote:
Look pretty reasonable to me, aside from a few nits.

facebook-github-bot-1:
@pvolok updated the pull request.

pkg/nuclide-ocaml/lib/OutlineProvider.js
pvolok: Yes, merlin reports items in the reversed order for some reason. I pushed `.reverse().map()` into the function to dedupe it.

pvolok:
Done.

nmote:
@facebook-github-bot import

facebook-github-bot-9:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3690807) to review internal test results.

nmote:
@pvolok Thank you! I'm also looking forward to using this when working in Flow.

nmote:
@pvolok I've been starting to do more work on Flow, and finally got my ocamlmerlin updated. I'm really enjoying the outline view. Thanks again for contributing this.

Issue #321, opened 2016-01-14
Nuclide Windows Support


mostafaeweda:
Nuclide is missing windows support, which causes many issues when users try stuff there.

SpencerSharkey:
I spent the past 20 minutes wondering why it wasn't working... D'oh!

SeananXiang:
can we  know  when nuclide will support windows perfect?

DimitrK:
Some info about that on the site or the repo would be so appreciated! People(including me) spent their time trying to figure out what's wrong with it.

Daniel15:
Basic functionality for remote projects appears to work on Windows. I'm using Windows 10 on the client side and Debian Linux on the server side. I can see the directory tree, edit remote files, and files are automatically reloaded if I edit them directly on the server. I haven't tested the more IDE-like functionality such as Flow/Hack integration, but I imagine these would be fine as they execute on the remote server rather than the local machine.

On the other hand, adding local project folders is totally broken. Nothing happens after clicking *Add Project Folder* and selecting a folder. The console has no errors, either.

whatsdis:
any updates on windows release and where can I test it out now on windows?

oizie:
This would be awesome to have in Windows. I have been trying to use without any major hacks but no luck so far... :/

ryancole:
Perhaps a run-down of what *does not* work on Windows, and why? Maybe that'll assist us in knowing why this can't have proper Windows support.

Daniel15:
Nuclide seems to run pretty well on Windows now. I've been using it for all
my day-to-day work at Facebook (I'm not on the Nuclide team, I'm just an
end user of Nuclide). I haven't encountered any major issues and for the
most part it seems quite usable.

Almost all of my work is done on a remote Linux server though, so I haven't
really tested Nuclide with local projects. There may be issues with
Hack/Flow for local projects as opposed to remote projects. Remote projects
run typechecking/linting remotely so it's less likely to encounter issues
with it.

ryancole:
@Daniel15 Ah, yea. I think the last time I tried out Nuclide (4-5 months ago) the Flow integration didn't work at all for local projects. Remote project makes more sense.

thetrompf:
I have managed to get the flow type checker up and running on windows.
[...]

PR #712, opened 2016-09-17
Which returns first result

zsol:
This fixes a problem on windows where `where.exe` can return multiple lines if the command is found multiple times on the `PATH`. This is especially painful since the recommended `npm install -g flow-bin` creates two `flow` binaries on by default so the entire flow support is broken, as mentioned in #321.

Also added tests.

### DIFF
--- a/pkg/commons-node/spec/which-spec.js
+++ b/pkg/commons-node/spec/which-spec.js
@@ -1,15 +1,24 @@
 'use babel';
 /* @flow */
 
+/*
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the LICENSE file in
+ * the root directory of this source tree.
+ */
+
 import which from '../which';
 
 describe('which', () => {
   let checkOutput: JasmineSpy;
-  let checkOutputReturn = {stdout: ''};
+  let checkOutputReturn: Object = null;
 
   beforeEach(() => {
+    checkOutputReturn = {stdout: ''};
     checkOutput = spyOn(require('../process'), 'checkOutput').andCallFake(() =>
-      checkOutputReturn
+      checkOutputReturn,
     );
   });
 
@@ -20,7 +29,7 @@ describe('which', () => {
   describe('on windows', () => {
     const real_platform: string = process.platform;
     const eol = '\r\n';
-    let os = require('os');
+    const os = require('os');
     const real_eol = os.EOL;
     beforeEach(() => {
       Object.defineProperty(process, 'platform', {value: 'win32'});
@@ -32,7 +41,7 @@ describe('which', () => {
     });
 
     it('calls where on Windows', () => {
-      let param: string = '';
+      const param: string = '';
       which(param);
       expect(checkOutput).toHaveBeenCalledWith('where', [param]);
     });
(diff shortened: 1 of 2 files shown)


facebook-github-bot-4:
By analyzing the blame information on this pull request, we identified @nmote to be a potential reviewer.

pkg/commons-node/spec/which-spec.js
```diff
@@ -0,0 +1,68 @@
...
+      Object.defineProperty(process, 'platform', {value: 'win32'});
+      os.EOL = eol;
+    });
+    afterEach(() => {
+      Object.defineProperty(process, 'platform', {value: real_platform});
+      os.EOL = real_eol;
+    });
+
+    it('calls where on Windows', () => {
+      let checkOutput: JasmineSpy = spyOn(require('../process'), 'checkOutput').andReturn('hello');
```
nmote: You're spying on `checkOutput` on every test and just changing what it returns. To remove boilerplate in each test, and properly unspy, I'd prefer to see, in a `beforeEach` in the top-level `describe`:

```
checkOutput = spyOn(require('../process'), 'checkOutput').andCallFake(() => checkOutputReturn);
```

with `checkOutput` and `checkOutputReturn` at the top.

Then unspy it in the `afterEach`.

Then in each test just set `checkOutputReturn`.

pkg/commons-node/spec/which-spec.js
```diff
@@ -0,0 +1,68 @@
...
+      let param: string = '';
+      which(param);
+      expect(checkOutput).toHaveBeenCalledWith('where', [param]);
+    });
+
+    it('returns the first match', () => {
+      waitsForPromise(async () => {
+        let returnValue = {stdout: 'hello' + os.EOL + 'hello.exe' + os.EOL};
```
nmote: It seems suspicious that the mocked return value is an object here and a simple string above. It looks like the actual return value isn't used above but still, it's easier to understand later if things are consistent.

facebook-github-bot-5:
@zsol updated the pull request - [view changes](https://github.com/facebook/nuclide/pull/712/files/63ff706f8c1d50a00a1223a091d1d0b230eb95b8..c11560e7075933827320593ce3a7455d81ecf0fc)

zsol:
Thanks @nmote, as you can tell I just started with jasmine :)

pkg/commons-node/spec/which-spec.js
```diff
@@ -0,0 +1,77 @@
+'use babel';
+/* @flow */
+
+import which from '../which';
+
+describe('which', () => {
+  let checkOutput: JasmineSpy;
+  let checkOutputReturn = {stdout: ''};
```
nmote: One more nit -- don't set a default value here, because if one test overrides it and another doesn't, the second one will end up with the value from the first test which is really confusing. Instead, set it to `null` here (one trick to satisfy Flow is to do something like `let checkOutputReturn: TypeAnnotation = (null: any)`), then set the default value at the top of the `beforeEach`.

facebook-github-bot-6:
@zsol updated the pull request - [view changes](https://github.com/facebook/nuclide/pull/712/files/c11560e7075933827320593ce3a7455d81ecf0fc..33027e7394a7d81d822eaa675bb44f0cfa949aba)

facebook-github-bot-1:
@zsol updated the pull request - [view changes](https://github.com/facebook/nuclide/pull/712/files/33027e7394a7d81d822eaa675bb44f0cfa949aba..2552a9b8cdee1a0f08ae6431b2a330f356dcea7b)

PR #737, opened 2016-09-23
Edited Debugger

zardra:


docs/_docs/features/debugger.md
```diff
@@ -32,148 +32,147 @@ project will be used to help illustrate the points described here.
...
 
-> If you stop your process and restart it again, make sure you press the `Refresh List` button since
-> the pid will have changed.
+> If you stop your process and restart it again, make sure you click the **Refresh List** button since
+> the PID will have changed.
 
-Once you attach to a process, you may see a confirmation of the attachment on the command-line for
-that process.
+Once you attach to a process, you may see a confirmation of the attachment in the command-line for
+that process. <!--[MAY OR WILL?]-->
```
JoelMarcey: At the time this was written, it wasn't always the case you would see a confirmation. However, I think theoretically, it should be "will". I won't block landing this because of that, but that can be changed in another pull request.

facebook-github-bot-4:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3917929).

docs/_docs/features/debugger.md
zardra: *facepalm* Completely forgot I had added that comment... Downside of doing my final review of the text in the browser.

JoelMarcey: It, of course, doesn't affect the rendering --- I would have pushed it --- but I think it should be "will", even if it doesn't always show up all the time currently. Documenting more for what should happen rather than what does :D

zardra: Let me change it to will and get rid of the comment...

facebook-github-bot-6:
@zardra updated the pull request - [view changes](https://github.com/facebook/nuclide/pull/737/files/ae8fd403c7382e29b6346cced41c5c12eea46b8e..0a35cdf08bbd1a3e8fad7fdbbb85990774a3f04d) - [changes since last import](https://github.com/facebook/nuclide/pull/737/files/ae8fd403c7382e29b6346cced41c5c12eea46b8e..0a35cdf08bbd1a3e8fad7fdbbb85990774a3f04d)

docs/_docs/features/debugger.md
zardra: Easy peasey... all fixed

Issue #747, opened 2016-09-26
Autocompletion fails with newer versions of SourceKitten


chika-kasymov:
### Issue and Steps to Reproduce
I installed last version of `Atom`, installed `Nuclide` package and it's dependencies, installed `SourceKitten`, reopened editor, but autocompletion not working. Maybe because of Swift 3.

### Versions
* Atom: 1.10.2
* Nuclide: 0.170.0
* Client OS: macOS Sierra 10.12.1
* SourceKitten: 0.14.1

modocache:
The Swift version shouldn't affect autocompletion. I'll look into this further. But first, could you confirm that you've built your project at least once via the Nuclide toolbar, as in the instructions here? https://nuclide.io/docs/languages/swift/#features__autocompletion

Autocompletion only works for projects that have been built at least once.

chika-kasymov:
@modocache previously I tried on raw swift file. After I created simple swift package and build with `Nuclide`. Package build successfully, but are there some errors with `SourceKitten` usage:
`An error occured when invoking SourceKitten. Unrecognized arguments: --compilerargs`

modocache:
Awesome, thanks. Something must have changed within SourceKitten. I'll take a look.

Keep in mind that this package doesn't currently support autocompletion on single Swift files. It only supports Swift packages built with the Swift package manager. :)

mominul:
@modocache Ya, now there is no `--compilerargs` option. That change was made by @keith. Please fix it soon, I'm facing that error!

modocache:
OK, I'll try to get to this soon. An alternative would be to send a pull request to SourceKitten that allows both kinds of arguments to be accepted. See the discussion on that issue here: https://github.com/jpsim/SourceKitten/pull/231

The onus is probably on me here to update it, which I'll try to do. Pull requests would also be really appreciated -- it should just be a matter of removing the `--` [here](https://github.com/facebook/nuclide/blob/d56a6a31fbe2ec635590e0a0987c413482aad749/pkg/nuclide-swift/lib/taskrunner/providers/SwiftPMAutocompletionProvider.js#L49) and confirming that everything works.

keith:
We could definitely allow this on the sourcekitten side. Ideally we would not add this just as a nop though. Is there any reason we can't remove that argument from nuclide and recommend sourcekitten upgrades?

modocache:
No reason at all! Just have to make a pull request. I'll try to do that soon. :)

keith:
[...]

PR #755, opened 2016-09-30
Remove SourceKitten compilerargs flag

keith:
This flag has been removed from SourceKitten in favor of just taking all
arguments after `--` as compiler arguments.

See https://github.com/jpsim/SourceKitten/pull/231 and https://github.com/facebook/nuclide/issues/747 for context


keith:
signed

zertosh:
@modocache - I don't know anything about this. It's your call. If you import it, I'll stamp it.

modocache:
Looks good to me, thanks @keith!

@zertosh How do I import it again? 😅

PR #756, opened 2016-09-30
Updated Linux commands and applied new formatting to other instructions.

zardra:
Updated the text to match the new formatting standard.

Changed the Linux install commands from 

$ wget https://atom.io/download/deb
$ mv deb atom-amd64.deb
$ dpkg -i atom-amd64.deb

to

$ sudo add-apt-repository ppa:webupd8team/atom
$ sudo apt-get update
$ sudo apt-get install atom

per Task #13640298.

### DIFF
--- a/docs/_docs/editor/setup.md
+++ b/docs/_docs/editor/setup.md
@@ -19,8 +19,7 @@ supported and stability is not guaranteed.
 
 ## Quick Install
 
-Assuming you have met all the prerequisites for your platform, culminating with **Atom 1.7.4 or
-greater** being installed, the easiest way to install Nuclide is either at the command line:
+Assuming you have met all the prerequisites for your platform, culminating with a version of Atom that is equal to or greater than that specified in [Atom's dependencies under "atom"](https://github.com/facebook/nuclide/blob/master/package.json) being installed, the easiest way to install Nuclide is either at the command line:
 
 ```bash
 $ apm install nuclide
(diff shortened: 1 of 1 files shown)


docs/_docs/editor/setup.md
```diff
@@ -19,8 +19,8 @@ supported and stability is not guaranteed.
 
 ## Quick Install
 
-Assuming you have met all the prerequisites for your platform, culminating with **Atom 1.7.4 or
-greater** being installed, the easiest way to install Nuclide is either at the command line:
+Assuming you have met all the prerequisites for your platform, culminating with *Atom 1.7.4 or
```
zertosh: Oh, we should totally fix these. The version we currently support is here: https://github.com/facebook/nuclide/blob/master/package.json#L114

docs/_docs/editor/setup.md
```diff
@@ -29,12 +29,11 @@ $ apm install nuclide
...
 
 1. Open Atom.
-2. Choose `Atom | Preferences` (on Linux this will be
-  `Edit | Preferences`) to bring up the `Settings` pane.
-3. In the `Settings` pane, click on `Install`.
-4. In the search box, type *Nuclide* and press the *Enter* key.
+2. Choose `Atom | Preferences` (on Linux this will be `Edit | Preferences`) to bring up the **Settings** tab.
+3. In the **Settings** tab, click on **Install**.
+4. In the search box, type "Nuclide" and press the `Enter` key.
 5. The package you are looking for is called `nuclide` and should be version `0.111.0` or greater.
```
zertosh: This reference to the version number can be removed. When you look for "nuclide", Nuclide is the first result. And the rest are unambiguously not Nuclide.

zardra:
Updated the Atom version number and removed the lines referring to the nuclide version.

docs/_docs/editor/setup.md
```diff
@@ -19,8 +19,8 @@ supported and stability is not guaranteed.
 
 ## Quick Install
 
-Assuming you have met all the prerequisites for your platform, culminating with **Atom 1.7.4 or
-greater** being installed, the easiest way to install Nuclide is either at the command line:
+Assuming you have met all the prerequisites for your platform, culminating with *Atom 1.10.2 or
```
JoelMarcey: Wanted to get your thoughts here...

Instead of pointing to a specific Atom version that will change over time, does it make sense to say something like:

"culminating with a version of Atom that is equal to or greater than that specified in [Atom's dependencies under "atom"](https://github.com/facebook/nuclide/blob/master/package.json)"

Or is that too much?

I am just thinking that way we don't have to continually update it.

zertosh: That's fine by me. Less places to have to keep updating :P

zardra:
Updated with the link

PR #761, opened 2016-10-04
Updated Debugger UI info

zardra:
Screenshots and information about Nuclide Debugger update to cover new interface.


JoelMarcey:
When this is reviewed and accepted -- will have to probably land this manually again given the binary files.

docs/_docs/features/debugger.md
```diff
@@ -6,7 +6,7 @@ permalink: /docs/features/debugger/
 ---
 
 One of the key features of Nuclide is its multiple-language debugging support. The Nuclide Debugger
-is provided within the familiar [Chrome DevTools](https://developer.chrome.com/devtools) interface.
+is provided with the familiar [Chrome DevTools](https://developer.chrome.com/devtools) interface.
```
JoelMarcey: Is this true any longer given the new interface? I know there is a setting to go back to the old chrome dev interface, but in general, the new interface is a bit different now, right?

JoelMarcey:
Love the new images. Just one question about us calling out Chrome Dev Tools....

docs/_docs/features/debugger.md
zardra: I changed  this from within to with because there is a button to click that pulls up the Chrome Dev Tools..... but I could see where just removing this sentence would be find since it's mentioned later on in the doc.

JoelMarcey: Ok - we can land this and think about whether we want to remove that sentence or not.

JoelMarcey:
landing internally...

Renamed 

`feature-debuggin-basics-stepping-controls.png` to `feature-debugger-basics-stepping-controls.png`

Issue #789, opened 2016-10-11
Asset and other links broken on doc pushes


JoelMarcey:
### Issue and Steps to Reproduce

Even when `baseurl` is "" or doesn't exist, a `/nuclide` is appended to our `site.baseurl` jekyll variable.

Push any diff (not even sure it has to be related to docs) to nuclide and watch our nuclide.io website freak out.
### Expected Behavior

The website layout should remain the same.
### Actual Behavior

Website freaks out.

![screenshot 2016-10-11 13 51 03](https://cloud.githubusercontent.com/assets/3757713/19291247/d18476e8-8fc8-11e6-81ed-b60f580617dd.png)

See example of appended `/nuclide` - https://gist.github.com/JoelMarcey/f71cc9b44041de272af9a589056184dd#file-gistfile1-txt-L9
### Additional Details
- Just keeping this here as a placeholder to make sure we don't forget it as I try to work around the issue.
- Working with GitHub support to see why this is happening.

JoelMarcey:
revert of the first attempt at temp fix: https://github.com/facebook/nuclide/commit/0a8c34b7075912115719d1756c3f6d048834f11b

JoelMarcey:
d73c78da6bc92c6bfcf49c2e8b334cff00e3612c is promising -- removing all instances of `baseurl` and `site.baseurl`.

Able to land docs diff with no issues.

Test some more tomorrow.

cc: @jamesgpearce

JoelMarcey:
Landed another docs diff today with https://github.com/facebook/nuclide/commit/d73c78da6bc92c6bfcf49c2e8b334cff00e3612c still in tact and no issues again. Looking good.

cc: @zardra

JoelMarcey:
https://github.com/facebook/nuclide/commit/a993b4eda4aad4868784ee3063ab7471234e4595 closed this

PR #793, opened 2016-10-12
Maybe baseurl is really the problem with the asset link fiasco

JoelMarcey:
So remove it all together and see what happens.

Ref #789


JoelMarcey:
closes #789

PR #800, opened 2016-10-13
Edited Building from Source doc

zardra:


docs/_docs/advanced-topics/building-from-source.md
```diff
@@ -18,17 +18,16 @@ those willing to live on the bleeding edge, you can install Nuclide from source.
 You must have the [general prerequisites](/docs/editor/setup#mac__prerequisites) installed. In
 addition, you must have
 
-- Xcode (for command line tools)
+- Xcode (for Command Line Tools)
 - **Node 5.10.0** or greater
```
JoelMarcey: Maybe we should do the same like we did with Atom and point to the package.json file so we do not have to keep up with version information. What do you think? Would do the same in the Linux part below too.

https://github.com/facebook/nuclide/blob/master/package.json#L114

docs/_docs/advanced-topics/building-from-source.md
```diff
@@ -79,8 +78,7 @@ addition, you must have **Node 5.10.0** installed as well.
```
JoelMarcey: Same here to point to https://github.com/facebook/nuclide/blob/master/package.json#L114

JoelMarcey:
but we can land that in a different diff

PR #811, opened 2016-10-15
Updating front page images (Fixes #807)

### LINKED ISSUE #807: Update feature screenshots on landing page
We should probably do at least one or more image updates on the `nuclide.io` landing page -- the debugger screenshot should be changed, for example.

The task runner should be added, I think as a main feature. We can either take one out or add it and another one (want an even n x 2 grid).

zardra:


JoelMarcey:
I can get the images aligned height and width -- but some then look stretchy. So, I am trying to put my CSS hat on to figure out how best to solve that problem.

JoelMarcey:
One option is that we take the current images on the landing page, re-screenshot them to exactly the same proportions and then use a specific `img.width` in the CSS and set `img.height` to `auto` or something.

JoelMarcey:
I got something that doesn't look too bad.... I think :)

Basically, in `_base.scss` force the object to fit in the container, even if we have to crop a bit.

```
.gridBlock {
  margin: 20px 0;
  padding: 2vh 0;

  .twoByGridBlock {
    padding: 0;

    img {
      margin-top: 6vh;
      width: 670px;
      height: 370px;
      object-fit: cover;
    }
```

screencapture-127-0-0-1-4000-1476733101829

JoelMarcey:
Then we can add some to `gridblocks.html` to check for the last picture, and if is an odd number (e.g., the 7th picture), then center it.

zardra:
The Task Runner one looks a bit odd... wonder if I can retake it to get at least the Task Runner toolbar into the shot.  It's still too bad that just resizing the image didn't do anything.

JoelMarcey:
Ok, to center an odd item, I added a `center` tag in the YAML. We would basically only put that if we have an odd number of images to display on the landing page. Then I said, if that `center` tag exists, we use this new `centerInGrid` styling that moves the margin of the gridblock over to the middle. It is 25% and not 50% because we are in a 2x2 grid styling.

So these are the changes you would make, where the `+` signs are.

> Note in `gridblock.html`, it is not a mandatory change. I just indented the `<img` tage for readability.

```
diff --git a/docs/_data/features.yml b/docs/_data/features.yml
index ecc333a..1f238b3 100644
--- a/docs/_data/features.yml
+++ b/docs/_data/features.yml
@@ -44,3 +44,4 @@
     noise. \\
     [**Read More**](/docs/features/working-sets/)
   image: images/docs/promo-working-sets.png
+  center: true
diff --git a/docs/_includes/content/items/gridblock.html b/docs/_includes/content/items/gridblock.html
index 8fb8331..d79b5f8 100644
--- a/docs/_includes/content/items/gridblock.html
+++ b/docs/_includes/content/items/gridblock.html
@@ -1,6 +1,6 @@
-<section id="fb_hhvm" class="featureBlock {{ include.gridtype }} fbossFontLight">
+<section id="oss_nuclide" class="featureBlock {{ include.gridtype }} {% if item.center %} centerInGrid {% endif %} fbossFontLight">
   {% if item.image %}
-  <img src="/static/{{ item.image }}" alt="{{ item.title }}" title="{{ item.title }}" />
+    <img src="/static/{{ item.image }}" alt="{{ item.title }}" title="{{ item.title }}" />
   {% endif %}
   <h3>{{ item.title }}</h3>
   {{ item.text | markdownify }}
diff --git a/docs/_sass/_base.scss b/docs/_sass/_base.scss
index 40b2117..1e3c46e 100644
--- a/docs/_sass/_base.scss
+++ b/docs/_sass/_base.scss
@@ -700,6 +700,10 @@ a.blockButton {
       width: 50%;
     }
 
+    .centerInGrid {
+      margin-left: 25%;
+    }
+
     .leftBlock {
       padding: 40px 20px;
     }
```

PR #818, opened 2016-10-20
Edited FAQ and Troubleshooting docs

zardra:
Removed Custom $PATH section from Troubleshooting since the Atom issue has been closed.


docs/_docs/help/troubleshooting.md
```diff
@@ -22,17 +22,15 @@ check out the following tips:
 
 *Node Version*
 
-Verify that you are using the correct node version by running:
+Use the command-line to verify that you are using Node version 5.10.0 or higher by running:
```
JoelMarcey: Does it make sense to be less specific on the version here and say like we do in another places...

> A version of Node that is equal to or greater than that specified in [“node” (under “engines”) in the package information](https://github.com/facebook/nuclide/blob/master/package.json)

zardra:
Updated the node language until we can either get something fancier working or get an easy install going.

PR #829, opened 2016-10-24
Replace Test Runner with Context View on home page

JoelMarcey:
The test runner needs some work. The context view is more full featured.
The latter currently will be of more use to the general Nuclide user at
this point. So swap them on the home page.


JoelMarcey:
screenshot 2016-10-23 18 42 58
screenshot 2016-10-23 18 42 36
screenshot 2016-10-23 18 42 22

JoelMarcey:
If there is another feature that should be showcased instead, let me know and we can change it. I just currently think the Test Runner should be removed until we give it a bit more love.

cc: @zardra

PR #845, opened 2016-10-28
Support flow error reporting for missing props

joarwilk:
Some flow messages have multiple error ranges, the flow diag tool should show the error highlight for all ranges.

This allows prop validation like this:
screen shot 2016-10-28 at 12 02 06

### DIFF
--- a/pkg/nuclide-flow/lib/FlowDiagnosticsProvider.js
+++ b/pkg/nuclide-flow/lib/FlowDiagnosticsProvider.js
@@ -70,13 +70,12 @@ function flowMessageToTrace(message: MessageComponent): Trace {
 }
 
 function flowMessageToDiagnosticMessages(diagnostic: Diagnostic) {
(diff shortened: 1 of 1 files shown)


nmote:
I understand the motivation here but I disagree that this is desirable behavior. Showing multiple error messages (a) duplicates information, (b) makes it so that it seems like there are more errors than there actually are, and (c) makes the number of errors inconsistent with the command line output, which is liable to confuse people.

Usually people want this because Flow reports the error in a confusing location. Instead of showing it at all locations, it would be better if Flow just reported it at the *best* location. In your example (based on a similar example I just put together), Flow reports the main location as that where the `props` property is defined. I agree that this is a confusing location. However, a better solution would be to have Flow report a different main error location. The Flow team has agreed to treat cases like this as bugs. I recommend opening an issue on Flow or searching for a similar one.

Anyway, thanks for putting up the PR. I appreciate the time spent even though I ultimately disagree that this is a step forward.

PR #862, opened 2016-11-03
Remove rogue + from delegate

JoelMarcey:


JoelMarcey:
screenshot 2016-11-03 10 11 58

zertosh:
Nope, that's correct. We updated to flow 0.34.0 and that's the syntax for [variance](https://flowtype.org/blog/2016/10/04/Property-Variance.html). You probably have an outdated `node_modules` so babel didn't strip out the `+`.

JoelMarcey:
Thanks!

Issue #673, opened 2016-08-23
Uncaught TypeError: Cannot read property 'ref' of undefined


wuqiuhao:
[Enter steps to reproduce below:]
I use Atom1.9.9 + nuclide edit React Native. I want to debug in Atom,but show this error.
1.  start React Native Packages
2. start Debugging
3. react-native run-ios

**Atom Version**: 1.9.9
**System**: Mac OS X 10.11.6
**Thrown From**: [nuclide](https://github.com/facebook/nuclide) package, v0.163.0
### Stack Trace

Uncaught TypeError: Cannot read property 'ref' of undefined

```
At /Users/wuqiuhao/.atom/packages/nuclide/pkg/nuclide-debugger-node/VendorLib/node-inspector/lib/InjectorClient.js:111

TypeError: Cannot read property 'ref' of undefined
  at InjectorClient.<anonymous> (/Users/wuqiuhao/.atom/packages/nuclide/pkg/nuclide-debugger-node/VendorLib/node-inspector/lib/InjectorClient.js:111:22)
  at /Users/wuqiuhao/.atom/packages/nuclide/pkg/nuclide-debugger-node/VendorLib/node-inspector/lib/DebuggerClient.js:121:7
  at Object.create.processResponse.value (/Users/wuqiuhao/.atom/packages/nuclide/pkg/nuclide-debugger-node/VendorLib/node-inspector/lib/callback.js:23:20)
  at Debugger._processResponse (/Users/wuqiuhao/.atom/packages/nuclide/pkg/nuclide-debugger-node/VendorLib/node-inspector/lib/debugger.js:95:21)
  at Protocol.execute (_debugger.js:120:14)
  at emitOne (events.js:90:13)
  at Socket.emit (events.js:182:7)
  at readableAddChunk (_stream_readable.js:153:18)
  at Socket.Readable.push (_stream_readable.js:111:10)
  at TCP.onread (net.js:531:20)

```

michaelswe:
I get the same error in Nuclide. Debugging in Chrome works fine.

joey3060:
get the same error in Nuclide

wuqiuhao:
I debugging in Chrome also have some problem！

digimane:
[...]

PR #865, opened 2016-11-07
Resolves issue facebook/nuclide#673

danmademe:
- Updated node-debugger to version from PR https://github.com/node-inspector/node-inspector/pull/932/files

Solved immediate issue.

NOTE: I'm not sure using node-debugger is wise anymore, the project looks abandoned. My guess is because node includes this now internally.
https://nodejs.org/api/debugger.html

### DIFF
--- a/pkg/nuclide-debugger-node-rpc/VendorLib/node-inspector/lib/InjectorClient.js
+++ b/pkg/nuclide-debugger-node-rpc/VendorLib/node-inspector/lib/InjectorClient.js
@@ -105,10 +105,10 @@ InjectorClient.prototype._findNMInScope = function(funcHandle, cb) {
         return prop.name == 'NativeModule';
       });
 
-      if (!NM.length)
-        error = new Error('No NativeModule in target scope');
+      if (!NM.length || !NM[0])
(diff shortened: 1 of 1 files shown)


danmademe:
Ok I filled out the CLA

johnislarry:
Thank you @danmademe!

Mohamed-Abo-El-Soud:
@danmademe You mean the nuclide debugger? I totally use it with react native because running the debugger from chrome makes javascript calls REALLY SLOW makes me unable to test without losing my sanity. Please don't kill this.

danmademe:
I don't work at Facebook. But it's not me that's gonna kill it it's the node-debugger package is now no longer maintained. Plus node has its own way of doing this. 

It's slow for you because your doing it in chrome. That's not an ideal way. The most ideal way would be to use the built in debugger in node and access it in the IDE. 

From memory the new method can be connected to via some web socket of sorts. So when someone works on a way of joining the two then your good to go. 

Also. Debuggers are supposed to be slow. That's just the way things are. If you hate the speed of debugging you should try debugging large scale Java applications. That is really really slow. JavaScript debugging is really fast by comparison.

PR #869, opened 2016-11-09
[Lint] Skip parsing "arc lint" output when there are no linters

ide:
When there are no linters (ex: no .arclint file), shelling out to `arc lint` gets nothing back as output, just an empty string for stdout. Trying to split and JSON-parse the empty string causes Nuclide to print out a warning that says `Error parsing \`arc lint\` JSON output`. Avoiding the parse when there are no linters avoids this warning.

### DIFF
--- a/pkg/nuclide-arcanist-rpc/lib/ArcanistService.js
+++ b/pkg/nuclide-arcanist-rpc/lib/ArcanistService.js
@@ -273,26 +273,29 @@ async function execArcLint(
   const result = await niceCheckOutput('arc', args, getArcExecOptions(cwd));
 
   const output: Map<string, Array<Object>> = new Map();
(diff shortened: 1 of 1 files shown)


nmote:
@hansonw since you are neck deep in this code right now could you review?

hansonw:
I actually have a pending change that touches this code. If you don't mind, I'll just incorporate this change into my diff to avoid conflicts :D

ide:
@hansonw go for it

hansonw:
Included your change in https://github.com/facebook/nuclide/commit/e466e158be1d6cf60020e421a73d9a6821083dc2#diff-39f6ff4e40573f868dbc45fc0501cf56R274 :) Thanks!

PR #882, opened 2016-11-13
Add several missing methods to atom flow typedefs

jmorrell:
Add some new atom flow defs

This adds a couple of type defs that I need for my atom plugin.

`npm run flow` produces errors for me, but does so on master as well... How should an external contributor be running flow to make sure we don't introduce new flow errors?

I assume these definitions will be moved to https://github.com/flowtype/flow-typed at some point, but understand that there are some issues with that. If there's anything I can do to help with that, let me know. I have a couple of weeks free.

### DIFF
--- a/pkg/nuclide-external-interfaces/1.0/atom.js
+++ b/pkg/nuclide-external-interfaces/1.0/atom.js
@@ -365,6 +365,7 @@ declare class atom$Pane {
   moveItemToPane(item: Object, pane: atom$Pane, index: number): void,
   destroyItem(item: Object): boolean,
   itemForURI(uri: string): Object,
+  getActiveEditor(): ?atom$TextEditor,
 
   // Event subscriptions.
   onDidAddItem(cb: (event: {item: Object, index: number}) => void): IDisposable,
@@ -660,6 +661,7 @@ declare class atom$TextEditor extends atom$Model {
   id: number,
   firstVisibleScreenRow: number,
   rowsPerPage: number,
+  element: atom$TextEditorElement,
 
   // Event Subscription
   onDidChange(callback: () => void): IDisposable,
@@ -735,6 +737,7 @@ declare class atom$TextEditor extends atom$Model {
       screenLine?: boolean,
     },
   ): atom$Point,
+  bufferRowForScreenRow(number): number,
   getVisibleRowRange(): ?[number, number],
 
   // Decorations
@@ -879,6 +882,7 @@ declare class atom$TextEditor extends atom$Model {
   tokenForBufferPosition(position: atom$Point | [?number, ?number]): atom$Token,
   onDidConflict(callback: () => void): IDisposable,
   serialize: () => mixed,
+  getElement(): atom$TextEditorElement,
 }
 
 /**
@@ -900,6 +904,7 @@ declare class atom$TextEditorComponent {
     linesClientRect?: {top: number, left: number, bottom: number, right: number},
   ): {top: number, left: number, bottom: number, right: number},
   invalidateBlockDecorationDimensions(decoration: atom$Decoration): void,
+  getVisibleRowRange: [?number, ?number],
 }
 
 /**
@@ -909,6 +914,7 @@ declare class atom$TextEditorComponent {
 declare class atom$TextEditorPresenter {
   startBlinkingCursors: () => void,
   stopBlinkingCursors(visible: boolean): void,
+  getVisibleRowRange: [?number, ?number],
 }
 
 /**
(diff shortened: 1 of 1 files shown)


nmote:
Unfortunately our OSS contribution workflow isn't very streamlined. A lot of us would like to improve it but we just haven't found the time to invest. The issue that you're probably referring to is that we have several instances of `require('fb-something')` where that module is only available internally. We do this for things like analytics: we want to collect data on how FB employees are using Nuclide, but we don't want to collect data on open source users. There are also a few places where we set defaults differently for employees vs other people.

Anyway, if that's all that Flow is warning about, don't worry. You can uncomment the `FlowFB` line in the `.flowconfig` and if we've been diligent, that should ignore all of those type errors.

We do intend to move these to flow-typed. The main issue is that we need to add the concept of namespaces to flow-typed. Right now it is only suitable for npm pacakge libdefs. We started some discussion on a [related issue](https://github.com/flowtype/flow-typed/issues/412). We are hoping to get to adding namespaces in the coming weeks but if you are interested in helping out we could use the extra hands. Feel free to weigh in on that issue.

Anyway I'm going to look at your actual PR now.

pkg/nuclide-external-interfaces/1.0/atom.js
```diff
@@ -365,6 +365,7 @@ declare class atom$Pane {
   moveItemToPane(item: Object, pane: atom$Pane, index: number): void,
   destroyItem(item: Object): boolean,
   itemForURI(uri: string): Object,
+  getActiveEditor(): ?atom$TextEditor,
```
nmote: This isn't a documented API. We try to limit the number of private APIs we include here. Can you explain why you need it, and why you can't make do with the public APIs?

pkg/nuclide-external-interfaces/1.0/atom.js
```diff
@@ -660,6 +661,7 @@ declare class atom$TextEditor extends atom$Model {
   id: number,
   firstVisibleScreenRow: number,
   rowsPerPage: number,
+  element: atom$TextEditorElement,
```
nmote: Also a private API. You shouldn't need this. Instead, use `atom.views.getView(editor)`

pkg/nuclide-external-interfaces/1.0/atom.js
```diff
@@ -879,6 +882,7 @@ declare class atom$TextEditor extends atom$Model {
   tokenForBufferPosition(position: atom$Point | [?number, ?number]): atom$Token,
   onDidConflict(callback: () => void): IDisposable,
   serialize: () => mixed,
+  getElement(): atom$TextEditorElement,
```
nmote: Also undocumented. See above for how to get the TextEditorElement

pkg/nuclide-external-interfaces/1.0/atom.js
```diff
@@ -735,6 +737,7 @@ declare class atom$TextEditor extends atom$Model {
       screenLine?: boolean,
     },
   ): atom$Point,
+  bufferRowForScreenRow(number): number,
```
nmote: This is undocumented as well. I don't see why though, it seems like a perfectly reasonable thing to have as part of the public API. I'll let it in since it seems reasonable and there's no obvious non-hacky way to accomplish this using documented APIs.

pkg/nuclide-external-interfaces/1.0/atom.js
```diff
@@ -900,6 +904,7 @@ declare class atom$TextEditorComponent {
     linesClientRect?: {top: number, left: number, bottom: number, right: number},
   ): {top: number, left: number, bottom: number, right: number},
   invalidateBlockDecorationDimensions(decoration: atom$Decoration): void,
+  getVisibleRowRange: [?number, ?number],
```
nmote: Well, we're already pretty far into undocumented territory here so I'll allow it

jmorrell:
Hmm... I was unaware undocumented APIs were considered private. I, perhaps naively, assumed that the public documentation was lacking and went straight to the code. 

- Does this come from the Atom team or the Nuclide team?
- While it makes sense to avoid using undocumented methods, I'm not sure leaving them untyped provides any great benefit? Perhaps that only makes sense once it's under flow-typed and can be pegged against atom's version

I'll review and see where I can simply use the documented methods 👍 or see if I can add to the Atom docs. Many of these I inherited from the module I forked. Thanks for the review!

nmote:
* I'm sure it's a mix of both "private" and "just undocumented." Either way, only the public docs give us a good way of guessing which parts of the API are more-or-less guaranteed to be stable.
* Leaving potentially unstable APIs out of our typedefs means that we can't use them (without doing convoluted tricks to escape Flow, which are obvious in code review). So it does accomplish the intended purpose.

Anyway, we aren't wholly against adding undocumented APIs here, we just would like to do it sparingly. If you get them added them to Atom's docs then we will definitely add them here.

jmorrell:
I was able to rewrite my module using only the documented APIs / existing flow annotations so I'm going to close the PR. Thanks for the review and guidance!

PR #890, opened 2016-11-15
Task Runner Debug buttons still had plug icon, updated to bug

zardra:
The Debug buttons for Buck and HHVM in the Task Runner still had the plug icon.  Not sure where nuclide-debugger-php/lib/main.js affects but updated it too.  Also not 100% certain I referenced the correct icon.


hansonw:
@matthewwithanm made proper "Nuclicons" for this. Thanks anyway!!

PR #891, opened 2016-11-15
Change RegExp dependency for url matching.

Pangoraw:
Hi,
The [current](https://www.npmjs.com/package/urlregexp) URL RegExp dependency doesn't find any url in text sample like `"http://example.com",` or if it does, it opens the URL `http://example.com",` which is not what we want.

So instead of using this library which includes quotes in its matching (I don't know if it is a relevant feature), I replaced it with this [one](https://www.npmjs.com/package/url-regex).

### DIFF
--- a/pkg/nuclide-url-hyperclick/lib/HyperclickProviderHelpers.js
+++ b/pkg/nuclide-url-hyperclick/lib/HyperclickProviderHelpers.js
@@ -30,7 +30,14 @@ export default class HyperclickProviderHelpers {
 
     urlregexp.lastIndex = 0;
 
-    const url = match[0];
+    let url = match[0];
+
+    // Stripping off quotes
+    if (url.includes('"'))
+      url = url.split('"')[0];
+    if (url.includes("'"))
+      url = url.split("'")[0];
+
     const index = match.index;
     const matchLength = url.length;
 


package.json
```diff
@@ -85,7 +85,7 @@
     "symbol-observable": "1.0.4",
     "temp": "0.8.3",
     "url-join": "1.1.0",
-    "urlregexp": "1.0.2",
+    "url-regex": "^4.0.0",
```
nmote: We do use shrinkwrap, but we like to pin the version here anyway so that only things we are deliberately changing change when we regenerate the shrinkwrap. Please just specify "4.0.0."

zertosh:
`url-regex` looks like it matches more things that may not be urls. I don't feel strongly either way since `nuclide-url-hyperclick` has a really low hyperclick priority. We can either add this lib in the correct way (no `^` and proper shrinkwraps), or strip quotes in `nuclide-url-hyperclick`.

nmote:
`urlregexp` already matches a ton of things that aren't likely to be URLs so I'm not too worried about that. As long as my inline comments are addressed I'm 👍.

Pangoraw:
Just commited [this](https://github.com/facebook/nuclide/pull/891/commits/758e1906be9a66308c947573156691e3d0afd4f8), however it looks like the generated shrinkwrap has changed for dependencies not linked to `url-regex`.

zertosh:
Found a very annoying bug in the regex while testing this. It captures the closing parens in `[Reason](https://facebook.github.io/reason/)`. So you end up going to `https://facebook.github.io/reason/)` - which, of course, doesn't work.

@Pangoraw how about just stripping the quotes off a match?

Pangoraw:
Alright, I reverted to the previously used regexp package and now, if a url is matched but contains quotes, only the part before the first quote is used. So now the package seems to work nicely for both `json` and `markdown` files.

PR #904, opened 2016-11-25
u[Nuclide] Fix bug in ReasonML / OCaml formatters.

jordwalke:
Summary:The `extname`e function returns `.rei`/`.mli`, not `re`/`mli`.

Test Plan: Code formatting now works.

Reviewers:

CC:


chenglou:
Oops sorry about that. Looks good!

PR #906, opened 2016-11-27
Handle null values in linters wrapped by the LinterAdapter

oleander:
The linter provided by [steelbrain/linter](https://github.com/steelbrain/linter/blob/v1/lib/linter-registry.js) allows for `null` values to be returned by a provider, while the current [LinterAdapter](https://github.com/facebook/nuclide/blob/master/pkg/nuclide-diagnostics-store/lib/LinterAdapter.js#L153) wrapping it, doesn't. Here's the relevant code in [steelbrain/linter](https://github.com/steelbrain/linter/blob/v1/lib/linter-registry.js#L65).

The [linter-eslint](https://github.com/AtomLinter/linter-eslint) package is currently returning `null` ([pull request/discussion](https://github.com/AtomLinter/linter-eslint/pull/703), [current implementation](https://github.com/AtomLinter/linter-eslint/blob/master/src/main.js#L177)).

Example usage:

```javascript
// ...
export function provideLinter() {
  return {
    // ...
    lint(editor) {
      // ...
      return eslint.lint(text).then(function(result) {
        if (isTooOld(result)) return null;
        return result;
      });
    }
  }
}
```

The current implementation of Nuclide throws the following error on `null` when using the latest version of `linter-eslint`.

```
atom/packages/nuclide/pkg/nuclide-diagnostics-store/lib/LinterAdapter.js:86 Uncaught (in promise) TypeError: Cannot read property 'Symbol(Symbol.iterator)' of null
  at linterMessagesToDiagnosticUpdate (atom/packages/nuclide/pkg/nuclide-diagnostics-store/lib/LinterAdapter.js:86:21)
  at atom/packages/nuclide/pkg/nuclide-diagnostics-store/lib/LinterAdapter.js:164:36
  at undefined.next (<anonymous>:null:null)
  at step (atom/packages/nuclide/node_modules/async-to-generator/async-to-generator.js:11:30)
  at atom/packages/nuclide/node_modules/async-to-generator/async-to-generator.js:21:13
```

This PR aims to solve this by checking the return value of a linter.


nmote:
In steelbrain/linter, what happens when a provider returns `null`, or a promise that resolves to `null`? Is it the same as it returning an empty set of lints?

As implemented, this PR means that if a provider returns some results, then on a subsequent call returns `null`, the previous results will remain visible. Is that intentional, or do you want to invalidate the current messages so that nothing is displayed?

hansonw:
@nmote The linter package does what you mentioned in its implementation here: https://github.com/steelbrain/linter/blob/v1/lib/linter-registry.js#L65 so I think that's reasonable. I'd imagine you would return empty to invalidate the current messages.

nmote:
Perfect, then 👍

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
```diff
@@ -149,7 +149,16 @@ export class LinterAdapter {
...
   async _runLint(editor: TextEditor): Promise<void> {
     if (this._enabled) {
-      const result = await this._requestSerializer.run(this._provider.lint(editor));
+      const maybe = this._provider.lint(editor);
+      if (!maybe) {
+        return;
+      }
+
+      const result = await this._requestSerializer.run(maybe);
+      if (!result) {
```
nmote: `RequestSerializer::run` never returns `null` or `undefined`. If the underlying promise resolves to `null` it will return an object with a `null` `result` field.

Instead you'll need to check `linterMessages` (below).

This change should also change the types in `nuclide-diagnostics-common/lib/main.js` to reflect the actual `lint` return type. Then Flow should guide you through the necessary changes.

oleander:
Thanks for your response guys. Before I continue, could you shed some light on these errors I get running `scripts/dev/test --no-version` on master.

Are they "by design", or did I miss a build step? I don't get any errors related to `LinterAdapter.js` before or after my commit.

```
☺  scripts/dev/test --no-version
Running lint_packages...
Finished lint_packages (0 seconds)
Running flow...
pkg/commons-node/passesGK.js:23
 23:     Gatekeeper = require('./fb-gatekeeper').Gatekeeper;
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb-gatekeeper. Required module not found

pkg/commons-node/process.js:436
436:   FB_INCLUDE_PATHS = require('./fb-config').FB_INCLUDE_PATHS;
                          ^^^^^^^^^^^^^^^^^^^^^^ ./fb-config. Required module not found

pkg/hyperclick/lib/main.js:29
 29:     const {overrideGoToDeclaration} = require('./fb/overrideGoToDeclaration');
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/overrideGoToDeclaration. Required module not found

pkg/nuclide-analytics/lib/track.js:20
 20:   module.exports = require('../fb/analytics');
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^ ../fb/analytics. Required module not found

pkg/nuclide-arcanist/lib/ArcBuildSystem.js:52
 52:       ArcToolbarModel = require('./fb/FbArcToolbarModel').FbArcToolbarModel;
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/FbArcToolbarModel. Required module not found

pkg/nuclide-blame-provider-hg/lib/HgBlameProvider.js:104
104:   const {getPhabricatorUrlForRevision} = require('./fb/FbHgBlameProvider');
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/FbHgBlameProvider. Required module not found

pkg/nuclide-clang-rpc/lib/ClangFlagsManager.js:76
 76:       _overrideIncludePath = require('./fb/custom-flags').overrideIncludePath;
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/custom-flags. Required module not found

pkg/nuclide-clang-rpc/lib/ClangServerManager.js:35
 35:       _getDefaultFlags = require('./fb/custom-flags').getDefaultFlags;
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/custom-flags. Required module not found

pkg/nuclide-clang-rpc/lib/find-clang-server-args.js:29
 29:       fbFindClangServerArgs = require('./fb/find-clang-server-args');
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/find-clang-server-args. Required module not found

pkg/nuclide-debugger-native/lib/LLDBLaunchAttachProvider.js:42
 42:       this._loadAction(require('./actions/fb-omActionUIProvider'));
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./actions/fb-omActionUIProvider. Required module not found

pkg/nuclide-debugger-php/lib/AttachProcessInfo.js:36
 36:       const services = require('./fb/services');
                            ^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/services. Required module not found

pkg/nuclide-diff-view/lib/DiffViewComponent.js:62
 62:       const {DiffViewPublishForm} = require('./fb/DiffViewPublishForm');
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/DiffViewPublishForm. Required module not found

pkg/nuclide-diff-view/lib/DiffViewComponent.js:77
 77:       const {DiffViewCreateForm} = require('./fb/DiffViewCreateForm');
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/DiffViewCreateForm. Required module not found

pkg/nuclide-diff-view/lib/RevisionTimelineNode.js:94
 94:       const diffUtils = require('../../commons-node/fb-vcs-utils.js');
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ../../commons-node/fb-vcs-utils.js. Required module not found

pkg/nuclide-diff-view/lib/redux/Epics.js:722
722:   Object.assign(module.exports, require('../fb/Epics'));
                                     ^^^^^^^^^^^^^^^^^^^^^^ ../fb/Epics. Required module not found

pkg/nuclide-flow/lib/FlowServiceWatcher.js:69
 69:     const reportButton = await require('./fb-report-crash').getButton();
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb-report-crash. Required module not found

pkg/nuclide-hg-repository-client/lib/HgRepositoryClient.js:81
 81:     const FbRevisionStatusCache = require('./fb/RevisionStatusCache');
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/RevisionStatusCache. Required module not found

pkg/nuclide-hg-rpc/lib/HgService.js:197
197:     refinements = require('./fb/config').primaryWatchSubscriptionRefinements;
                       ^^^^^^^^^^^^^^^^^^^^^^ ./fb/config. Required module not found

pkg/nuclide-nux/lib/NuxStore.js:44
 44:       NuxBackendCache = require('./fb-NuxCache').NuxCache;
                             ^^^^^^^^^^^^^^^^^^^^^^^^ ./fb-NuxCache. Required module not found

pkg/nuclide-python-rpc/lib/JediServerManager.js:27
 27:     overrides = await require('./fb/find-jedi-server-args')(src);
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/find-jedi-server-args. Required module not found

pkg/nuclide-python-rpc/lib/PythonService.js:100
100:     const overridePath = require('./fb/find-formatter-path')();
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/find-formatter-path. Required module not found

pkg/nuclide-python-rpc/lib/PythonService.js:184
184:     result = await require('./fb/run-flake8')(src, contents, configPath);
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^ ./fb/run-flake8. Required module not found

pkg/nuclide-remote-projects/lib/connection-profile-utils.js:160
160:     defaultConfigGetter = require('./fb/config');
                               ^^^^^^^^^^^^^^^^^^^^^^ ./fb/config. Required module not found


Found 23 errors
```

```
☃  atom --version                                                                                                                                [oleander@laptop][14:07:39]
Atom    : 1.12.5
Electron: 1.3.9
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

```
☺  git rev-parse HEAD # master
abddcc12fd9d9f5e61ffde1be95008b0c3b6f9db
```

```
☺  flow version
Flow, a static type checker for JavaScript, version 0.35.0
```

nmote:
That's our fault, and frankly it's something I'm really unhappy about. We have various places where for the sake of expedience we've added some facebook-internal functionality into these `fb-` prefixed files, which ship with our internal version but do not get synced to GitHub. This means that when running Flow on the GitHub copy, you will see these errors.

I don't have a great solution for you, to be honest. Probably just run Flow yourself and ignore these errors. And for running tests, just run the tests for the package you are working on.

Fixing this properly will take a concerted effort on our part. A lot of us want to do it but we just haven't had the time. In the meantime if you have ideas for how to ease the pain I would be more than happy to review pull requests (maybe adding options to the script or something). Feel free to open issues to discuss ideas and CC me so I get to them promptly.

hansonw:
For the time being a lot of these fb-only requires are preceded by a `// $FlowFB` so uncommenting this line https://github.com/facebook/nuclide/blob/master/.flowconfig#L31 in the .flowconfig may reduce the noise.

nmote:
But, I don't think we've been diligent about adding those comments so it might not get all of them. And be careful not to include the config change in your PR :/

oleander:
@nmote Now everything passed.

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
```diff
@@ -148,35 +148,49 @@ export class LinterAdapter {
...
-          });
-          this._onDestroyDisposables.set(buffer, disposable);
-        }
-
-        const linterMessages = result.result;
-        const diagnosticUpdate = linterMessagesToDiagnosticUpdate(
-          editor.getPath(),
-          linterMessages, this._provider.providerName || this._provider.name,
-        );
+    if (!this._enabled) {
```
nmote: Good call with the early returns here. It reads much better.

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
```diff
@@ -148,35 +148,49 @@ export class LinterAdapter {
...
+    if (result.status !== 'success') {
+      return;
+    }
+
+    const buffer = editor.getBuffer();
+    if (buffer.isDestroyed()) {
+      return;
+    }
+
+    if (this._provider.invalidateOnClose && !this._onDestroyDisposables.has(buffer)) {
```
nmote: Any particular reason why this block happens before the `linterMessages` null check? Or did you just choose the order arbitrarily?

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
```diff
@@ -148,35 +148,49 @@ export class LinterAdapter {
...
-        const diagnosticUpdate = linterMessagesToDiagnosticUpdate(
-          editor.getPath(),
-          linterMessages, this._provider.providerName || this._provider.name,
-        );
+    if (!this._enabled) {
+      return;
+    }
+
+    const maybe = this._provider.lint(editor);
+    if (!maybe) {
```
nmote: We prefer to do null/undefined checks explicitly: `if (maybe == null)`. Better to write what you actually mean even though it's a few more characters, rather than relying on truthiness which has some odd quirks.

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
```diff
@@ -148,35 +148,49 @@ export class LinterAdapter {
...
-        }
-      }
+        this._onDestroyDisposables.delete(buffer);
+        disposable.dispose();
+      });
+      this._onDestroyDisposables.set(buffer, disposable);
+    }
+
+    const linterMessages = result.result;
+    if (!linterMessages) {
```
nmote: same here: `linterMessages == null`

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
nmote: Yeah, that's the one exception we make in our config. It's basically due to Flow's `?something` being shorthand for `something | null | undefined`. Since it's used so pervasively we basically have to conflate `null` and `undefined`. Oh well.

pkg/nuclide-diagnostics-store/lib/LinterAdapter.js
oleander: Tried to make as small changes as possible until the spec passed. The `linterMessages = result.result` row was placed below the block so kept it there, just in case. Do you want me to move it?

The code would look more symmetrical if the guard was moved above the block, so I don't mind moving it.

nmote: I don't really care either way. I don't think it's important. I just wanted to make sure there wasn't some subtlety I was overlooking.

oleander:
Fixed all bugs™

nmote:
Thanks, LGTM

PR #909, opened 2016-11-30
Navigation Stack in the FAQ and keyboard shortcuts

JoelMarcey:

PR #919, opened 2016-12-07
Swap navigation forward/backward keys on PCs

phord:
Windows and Linux used 'Ctrl-<' to go forwards and 'Ctrl->' to
go backwards in history.  This is opposite the expected direction.
Swap these two key definitions.

The Mac/OS X key definitions were correct, but the documentation also
had them backwards.  Swap the key definition descriptions in the
documentation.

A remaining issue appears to exist which blocks Ctrl-< since it is
bound to a different action by Atom-Core.  This change doesn't correct
that because it's a different issue and it's not obvious how it should
be fixed in Nuclide.

Fixup for #909

### DIFF
--- a/docs/_docs/editor/keyboard-shortcuts.md
+++ b/docs/_docs/editor/keyboard-shortcuts.md
@@ -166,8 +166,8 @@ These are keyboard shortcuts with respect to navigation within files, etc.
 
 | Key (macOS) | Key (Linux) | Command | Description |
 |-----------|-------------|---------|-------------|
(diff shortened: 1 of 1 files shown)


phord:
I see this was fixed separately in 50c638ead9c55895860c58f3730258f64281187a, but the documentation wasn't updated in that commit.  Should be easy enough to merge, still.

JoelMarcey:
@phord Hi Phil! Could rebase this pull request against that commit, so we don't have any merge conflicts. I know, simple change, but it would just be cleaner that way; otherwise, we have to manually resolve conflicts, etc. Thanks!

PR #924, opened 2016-12-12
keymaps: Ctrl instead of Cmd on Linux/Windows

phord:
* Ctrl-i for context-view-toggle
  * Ctrl-Shift-{a,y} for debugger{,-launch-attach}

This matches the current documentation already.


hansonw:
Cool, thanks! Have you verified that these don't conflict with any existing Atom default keybinds?

phord:
I checked some of them, but I'm not certain I checked all.  I did verify that the ones listed in the documentation do match what's in this change.

hansonw:
Thanks for the PR! in the future please make sure they don't have trailing commas though (142c066). It seems we don't have any lint protection against this yet :(

PR #928, opened 2016-12-13
New images for the Nuclide landing page

zardra:


JoelMarcey:
Thanks for going through the trouble of doing this.

PR #943, opened 2016-12-18
Drop `cheerio`.

wtgtybhertgeghgtwtg:
Please, correct me if I am wrong, but `nuclide` does not seem to use it anywhere.


zertosh:
Sorry we use this internally :(

wtgtybhertgeghgtwtg:
Sorry about that.  Thank you for the quick response.

PR #955, opened 2016-12-31
Flow autocomplete: ignore prefix unless it's id.

pvolok:
I'm adding an autocomplete for GraphQL queries in Flow. In the following example Nuclide filters out all suggestions because the prefix is `" { "`.

```js
Relay.QL`query { <HERE> }`;
```

As far as I understand prefix filtering is only useful if prefix is an identifier.

### DIFF
--- a/pkg/nuclide-flow/lib/FlowAutocompleteProvider.js
+++ b/pkg/nuclide-flow/lib/FlowAutocompleteProvider.js
@@ -19,6 +19,8 @@ import passesGK from '../../commons-node/passesGK';
 
 import {getFlowServiceByNuclideUri} from './FlowServiceFactory';
 
+const ID_REGEX = /^[a-zA-Z_$][0-9a-zA-Z_$]*$/;
+
 export default class FlowAutocompleteProvider {
   _cacher: AutocompleteCacher<?Array<atom$AutocompleteSuggestion>>;
   constructor() {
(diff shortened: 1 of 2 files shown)


pvolok:
Also, with this change we now have autocompletion for objects:

```js
({ <HERE> });
({ a: <HERE> });
```

pkg/nuclide-flow/lib/FlowAutocompleteProvider.js
```diff
@@ -134,9 +134,9 @@ function updateResults(
 }
 
 function getReplacementPrefix(originalPrefix: string): string {
-  // If it is just whitespace and punctuation, ignore it (this keeps us
-  // from eating leading dots).
-  return /^[\s.]*$/.test(originalPrefix) ? '' : originalPrefix;
+  // Ignore prefix unless it's an identifier (this keeps us form eating leading
```
nmote: typo (from/form)

pkg/nuclide-flow/lib/FlowAutocompleteProvider.js
```diff
@@ -134,9 +134,9 @@ function updateResults(
...
+  // dots, colons, etc).
+  return /^[a-zA-Z0-9_$]*$/.test(originalPrefix) ? originalPrefix : '';
```
nmote: I've inlined an identifier regex in `shouldFilter` in this same file -- could you pull that out into a constant and use it here? Notably, what you have here allows numbers at the beginning of an identifier which is incorrect (not a big issue in this case but why not do it right).

pvolok:
I added one test to check that we don't filter when prefix is not an id, otherwise is tested.

nmote:
Looks good, thanks!

PR #988, opened 2017-01-18
Do not use jQuery when getting versions from package.json

JoelMarcey:
- Use classes instead of ids for the placeholders since getElementById may only get the first one on a page
- Also put in footer to ensure DOM is loaded
- And put script in its own js file


JoelMarcey:
Thanks @zertosh for the non-jquery suggestion.

PR #1004, opened 2017-01-26
Fix nuclide-adb-logcat parsing

fcFn:
There is an additional CR before CRLF in log lines, which prevents proper parsing of log lines from  adb (version 1.0.36).

Also, some React errors have `unknown:React` tag and will not get parsed using the default configuration:

```
[ 01-26 13:20:34.550  3053: 3053 E/unknown:React ]
Exception in native call
com.facebook.react.devsupport.DebugServerException:
...
```

### DIFF
--- a/pkg/nuclide-adb-logcat/lib/createProcessStream.js
+++ b/pkg/nuclide-adb-logcat/lib/createProcessStream.js
@@ -58,7 +58,7 @@ export function createProcessStream(): Observable<string> {
   )
     // Only get the text from stdout.
     .filter(event => event.kind === 'stdout')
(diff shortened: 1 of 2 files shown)


matthewwithanm:
Thanks for the PR @fcFn! The updated expression is going to require the `unknown:` though, right? So it's not going to match "React". Maybe just change it to `unknown:React|React`?

fcFn:
It will match both "unknown:React" and "React", you can check it here: https://repl.it/FWp4.

matthewwithanm:
derp, sorry. of course it will. thanks!

PR #1005, opened 2017-01-26
Updated notifyRevisionStatus's success pop-up to be dismissable

zardra:
Added `dismissable: true` to `notifyRevisionStatus` so that Publish to Phabriactor dialog remains until user dismisses it.

### DIFF
--- a/pkg/nuclide-diff-view/lib/utils.js
+++ b/pkg/nuclide-diff-view/lib/utils.js
@@ -519,6 +519,7 @@ function notifyRevisionStatus(
       text: 'Open in Phabricator',
     }],
     nativeFriendly: true,
+    dismissable: true,
   });
 }
 


mostafaeweda:
@zardra that was the case initially, but was removed in response to user feedback - please revise https://github.com/facebook/nuclide/commit/89e4838e026e0474e74caacbf2dd1b0474666828

JoelMarcey:
@mostafaeweda Did you mean "revert" that commit?

cc @jgebhardt 

Could we have a nuclide setting that can be toggled with respect to dialogs (or just this dialog) as a happy medium? Or is that too granular of a setting?

mostafaeweda:
Now that the console has clickable links (came in the last release) to go to the created/updated revision - I'd say it's totally fine if the dialog goes away, because you can use the console to go there.
However, a config option is also fine with me.  My point is: 89e4838 was done in response to user feedback, so we shouldn't just revert it.

PR #1006, opened 2017-01-28
Fix adb-logcat metadata regex for 5-digit decTID

fcFn:
If decTID is 5-digit (which doesn't happen very often), metadata will not be parsed:
```
[ 01-28 14:59:27.633  3211:15223 I/ReactNativeJS ]
```


matthewwithanm:
@fcFn Thanks so much for doing this! Is there any chance we could get a tiny test to make sure we don't regress?

fcFn:
Will updating the formats and associated tid test be enough? 

Or should I add another test just for 5-digit tid?

matthewwithanm:
@fcFn Thanks!!

PR #1012, opened 2017-02-02
Add support for project-wide C++ compilation flags for libclang

### LINKED ISSUE #916: Clang system headers are not included by default
### Issue and Steps to Reproduce
I get errors reported in diagnostics that I do not see when I compile the file in the shell using the command from my compile_commands.json.  Is there some what to verify the compiler command line being used?  I checked in the /tmp/nuclide-*-logs/nuclide.log, but the details are not there.  I see some results in /tmp/nuclide*/clang/, but still no hint as to what the used switches are.

### Expected Behavior
Clang compile should use the commands found in compile_commands.json at the top of the project.  Errors reported to Diagnostics should match what I see on the commandline console.

### Actual Behavior
Different errors are reported which suggest the wrong --std switch is used, wrong include files are found, etc.

### Versions
* Atom: 1.12.6
* Nuclide: 0.186.0
* Client OS: Ubuntu 16.10
* Server OS (optional): N/A
* libclang: /usr/lib/llvm-3.8/lib/libclang.so

### Additional Details
* Installed packages (`apm ls --installed`):
Community Packages (12) /home/phord/.atom/packages
├── haskell-grammar@0.4.0
├── highlight-selected@0.11.2
├── language-babel@2.48.5
├── language-ini@1.17.0
├── language-ocaml@1.1.2
├── language-swift@0.5.0
├── language-thrift@1.0.2
├── merge-conflicts@1.4.4
├── nuclide@0.186.0
├── nuclide-format-js@0.0.36
├── sort-lines@0.14.0
└── tool-bar@1.0.1

└── (empty)

Outline seems to work now, but autocomplete and "goto" do not seem right at all.
In Nuclide-clang I have all the checkmarks unchecked:
   [ ]  Enable diagnostics with default flags
   [ ]  Enable default compilation flags

*Edit: add libclang.so version*

AlexTalks:
Close #916 

This adds support for providing a `.nuclide_clang_config.json` file in the project root, in which compilation flags and include paths needed for libclang (such as system header include paths) can be provided.  In this file can be provided `extra_compiler_flags` (flags to include for each translation unit), and  `ignored_compiler_flags` (flags to filter out for each translation unit).


AlexTalks:
@hansonw I figure you're the right person to notify about this?

hansonw:
Nice, this is cool! I'm a bit reluctant to add a one-off configuration file for this (ideally Nuclide should have one definitive file for project-level configuration), but since we don't yet have a plan for that I'm happy to take this for now. Just keep in mind that if we ever create a `.nucliderc` or similar we will probably migrate this over :)

cc @a20012251, this may be useful for what you're working on too

I'll find some time to review the code in detail, but some bikeshedding first ;)

- Maybe we should make the file 'hidden', as is the standard for most configuration files. Perhaps also make the naming a bit more generic, i.e. `.nuclide_clang_config.json`.
- I think it's a bit redundant to have both extra flags and extra includes, when one is a special case of the other. Plus this adds ambiguity in terms of the ordering. Can we just let people add the `-isystem` to their flags manually to simplify the API?
- `copts` and `nocopts` are a bit cryptic. How about `extra_flags` and `exclude_flags`?

So in summary the file would look like `{"extra_flags": [...], "exclude_flags": [...]}`.

AlexTalks:
@hansonw sure!  sounds reasonable about moving to a standard `.nucliderc` at some point.

As for the naming, I tried to follow conventions set by Buck and Bazel, though Buck doesn't have the `nocopts` feature so I stuck with Bazel's naming.  In keeping more in line with Buck's naming however, I could just use `{ "compiler_flags": [..], "exclude_compiler_flags": [..] }` if that makes more sense.

pkg/nuclide-clang-rpc/lib/ClangFlagsManager.js
```diff
@@ -200,6 +211,33 @@ export default class ClangFlagsManager {
...
+    }
+
+    if (projectFlags == null) {
+      return originalFlags;
+    }
+
+    const extraCompilerFlags = projectFlags.extraCompilerFlags.flags || [];
+    const ignoredCompilerFlags = projectFlags.ignoredCompilerFlags.flags || [];
+
+    let finalFlags = originalFlags.slice(0);
```
hansonw: Nitpick: the slice is unnecessary. we can write this block in the functional style:

```
return originalFlags
  .filter(flag => ignoredCompilerFlags.indexOf(flag) === -1)
  .concat(extraCompilerFlags);
```

pkg/nuclide-clang-rpc/lib/ClangFlagsManager.js
```diff
@@ -200,6 +211,33 @@ export default class ClangFlagsManager {
...
+    if (projectFlagsDir != null) {
+      const projectFlagsFile = nuclideUri.join(projectFlagsDir, PROJECT_CLANG_FLAGS_FILE);
+      projectFlags = await this._loadProjectCompilerFlags(projectFlagsFile);
+    }
+
+    if (projectFlags == null) {
+      return originalFlags;
+    }
+
+    const extraCompilerFlags = projectFlags.extraCompilerFlags.flags || [];
```
hansonw: After changing these to plain arrays the `|| []` should be unnecessary

hansonw:
Actually I see in your code that you already used `extraCompilerFlags` and `ignoredCompilerFlags`.. I like those, so let's be consistent? `{"extra_compiler_flags": [], "ignored_compiler_flags": []}` then

pkg/nuclide-clang-rpc/lib/ClangFlagsManager.js
```diff
@@ -266,6 +304,34 @@ export default class ClangFlagsManager {
...
+      return this._clangProjectFlags;
+    }
+
+    let result = null;
+    try {
+      const contents = await fsPromise.readFile(flagsFile, 'utf8');
+      const data = JSON.parse(contents);
+      invariant(data instanceof Object);
+      const {copts, nocopts, includes} = data;
+      const extraCompilerFlags = {flags: [], rawData: null, flagsFile};
```
hansonw: As mentioned above, I'd just leave these as the raw arrays.
Also it's probably a good idea to make these fields optional, so make sure you handle the case where they're null.

pkg/nuclide-clang-rpc/lib/ClangFlagsManager.js
```diff
@@ -266,6 +304,34 @@ export default class ClangFlagsManager {
     return null;
   }
 
+  async _loadProjectCompilerFlags(flagsFile: string): Promise<?ClangProjectFlags> {
+    if (this._clangProjectFlags != null) {
```
hansonw: As mentioned above, you need to look this up by `flagsFile`. And you need to cache the promise- see the `getFlagsForSrc` cached vs impl pattern above

pkg/nuclide-clang-rpc/spec/ClangFlagsManager-spec.js
```diff
@@ -233,6 +243,37 @@ describe('ClangFlagsManager', () => {
     });
   });
 
+  it('gets project flags in addition to compilation database flags', () => {
+    waitsForPromise(async () => {
+      const sampleJsonFile = nuclideUri.join(__dirname, 'fixtures',
+        'nuclide_clang_flags.json.sample');
+      const tmpJsonFile = nuclideUri.join(__dirname, 'fixtures', 'nuclide_clang_flags.json');
```
hansonw: Let's use a real temporary directory for this test. We have a `copyFixture` utility (see https://github.com/facebook/nuclide/blob/648d770f3f58794dbdcfa1b0d9ccf6ae484e1581/spec/clang-refactor-integration-spec.js#L25), so let's copy out the fixtures directory. The temporary directory will be automatically cleaned up so no need for the manual `afterEach` either :)

hansonw (changes requested):
Looks good overall!

AlexTalks:
@hansonw thanks for the feedback!  will fix up your comments and push another commit.

pkg/nuclide-clang-rpc/lib/ClangFlagsManager.js
```diff
@@ -90,19 +97,22 @@ export default class ClangFlagsManager {
   _compilationDatabases: Map<string, Map<string, ClangFlags>>;
   _realpathCache: Object;
   _pathToFlags: Map<string, Promise<?ClangFlags>>;
+  _clangProjectFlags: ?ClangProjectFlags;
```
AlexTalks: good to know.  I didn't realize how that worked, but of course that makes sense.

AlexTalks:
@hansonw added the commit.  I'm not sure if @facebook-github-bot expects us to have amended our earlier commits, or have pushed a new commit?  (Also I somewhat assume that arcanist actually squashes the PR before merge or something)

hansonw:
Good stuff, thanks! Yeah our GitHub -> Phabricator flow will squash all your commits and close this PR. Just going to pass it through internal review first

a20012251:
very nice :) We might use it for our project

janosimas:
Is it already in last nuclide release?
any docs or example on how to crete the config file?

hansonw:
Yes this is in v0.204! Check the PR for docs (the C++ docs section was updated)

janosimas:
Hi, I did some tests and couldn't get it to work.

I included -stdlib=libc++ in .nuclide_clang_config.json but the diagnostic still cant find the stdlib.
any tips?

AlexTalks:
@janosimas an easy way to check it to see the running instances of `clang_server.py` and see what flags it is using for each translation unit.  For me I needed to make sure that the actual paths of the system header files were passed in through `-isystem` include flags.

janosimas:
@AlexTalks I'm doing some tests here but if I use the flags of the running instance of  the clang_server.py I can compile the file with no warning or error.
Is the clang_server.py behavior different from the clang?

hansonw:
@janosimas have you seen https://github.com/facebook/nuclide/issues/916? libclang seems to be unaware of the default system header location, so you have to extract them via `clang -### -c <file>`

janosimas:
After some sweating it worked, for some reason using -isystem for the gcc stdlib didn't work.
I has to install clang stdlib, set -stdlib=libc++ and isystem clangs include folders.

PR #1016, opened 2017-02-06
[React Inspector] Replace vendored React DevTools with npm package

gaearon:
This would make it easier for us to ship fixes and new features in long term. The npm package is in `react-devtools` repo and built from the same Electron shell you are currently vendoring.

As a bonus, the npm package contains the optimized React DevTools build. The currently vendored version uses a slower DEV build of React.

I tested that this worked locally by changing my Nuclide installation in place but I haven’t tried to rebuild the open source version. Let me know if I missed something!


matthewwithanm:
Thanks @gaearon! A couple of things:

- Dependencies need to go in the root package.json (not individual packages'). I was just going to fix this myself (and regenerate the shrinkwrap files) but…
- The react-devtools package is pulling in [electron-prebuilt](https://github.com/electron-userland/electron-prebuilt). That's a pretty huge (unnecessary) dependency, and it has a postinstall script, which we won't run. @zertosh suggests publishing a separate `react-devtools-client` package, that we and `react-devtools` can depend on.

Thoughts?

gaearon:
cc @jaredly, do you have a preference for how we package things? This is indeed an unfortunate limitation.

gaearon:
(Your plan with a small package just for the client sounds good to me.)

zertosh:
@gaearon, your changes to add the dep should look like this:

```diff
diff --git a/npm-shrinkwrap.json b/npm-shrinkwrap.json
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1837,6 +1837,11 @@
       "from": "react-addons-update@15.3.1",
       "resolved": "https://registry.npmjs.org/react-addons-update/-/react-addons-update-15.3.1.tgz"
     },
+    "react-devtools-core": {
+      "version": "1.0.6",
+      "from": "react-devtools-core@1.0.6",
+      "resolved": "https://registry.npmjs.org/react-devtools-core/-/react-devtools-core-1.0.6.tgz"
+    },
     "react-dom": {
       "version": "15.3.1",
       "from": "react-dom@15.3.1",
diff --git a/npm-shrinkwrap.production.json b/npm-shrinkwrap.production.json
--- a/npm-shrinkwrap.production.json
+++ b/npm-shrinkwrap.production.json
@@ -950,6 +950,11 @@
       "from": "react-addons-update@15.3.1",
       "resolved": "https://registry.npmjs.org/react-addons-update/-/react-addons-update-15.3.1.tgz"
     },
+    "react-devtools-core": {
+      "version": "1.0.6",
+      "from": "react-devtools-core@1.0.6",
+      "resolved": "https://registry.npmjs.org/react-devtools-core/-/react-devtools-core-1.0.6.tgz"
+    },
     "react-dom": {
       "version": "15.3.1",
       "from": "react-dom@15.3.1",
diff --git a/package.json b/package.json
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "nullthrows": "1.0.0",
     "plist": "2.0.1",
     "react": "15.3.1",
+    "react-devtools-core": "1.0.6",
     "react-dom": "15.3.1",
     "react-for-atom": "15.3.1-0",
     "redux": "3.6.0",
```

(https://github.com/facebook/nuclide/wiki/Adding-a-New-Package-Dependency)


We're planning on bumping `ws` to v2.0.0 soon (right @hansonw?), what's your timeline for updating `react-devtools-core`?

gaearon:
>what's your timeline for updating react-devtools-core?

Whenever convenient, only a single file uses it.

gaearon:
Pushed changes to the shrinkwrap.

package.json
```diff
@@ -69,6 +69,7 @@
     "nullthrows": "1.0.0",
     "plist": "2.0.1",
     "react": "15.3.1",
+    "react-devtools-core": "^1.0.6",
```
zertosh: Everything looks great, except this caret... but I can fix this on import.

gaearon: Thanks!

matthewwithanm:
@zertosh @gaearon Thanks you two, this looks great!! ❤️

Issue #998, opened 2017-01-24
Diagnostics panel is cutting Flow errors description


montogeek:
### Issue and Steps to Reproduce
I am using Nuclide with Flow, when you open the Diagnostics panel, the Description column is cutting Flow errors, like this:
![image](https://cloud.githubusercontent.com/assets/1002461/22249320/87cee35e-e210-11e6-83f7-f852fae5eef8.png)
This also happen when you select "Show only diagnostics for this file":
![image](https://cloud.githubusercontent.com/assets/1002461/22249326/96de602c-e210-11e6-86a6-3eabd83b1b1a.png)

But the inline errors show the error message fully:
![image](https://cloud.githubusercontent.com/assets/1002461/22249345/a6922cec-e210-11e6-9082-2a3d4fb2e708.png)

There is a lot of space to show the full error message in the Description column

I could work on its fix

### Expected Behavior
Description column in Diagnostic panel should show full Flow error message

### Actual Behavior
Description column in Diagnostic panel is cutting Flow error message

### Versions
* Atom: 1.13.0
* Nuclide: 201
* Client OS: OSX Sierra

### Additional Details
* Installed packages (`apm ls --installed`):
<details>
Community Packages (39) /Users/montogeek/.atom/packages
├── Zen@0.16.4
├── atom-material-ui@1.3.9
├── docblockr@0.9.1
├── duotone-dark-red-syntax@0.3.2
├── editorconfig@2.2.0
├── emmet@2.4.3
├── enzyme-snippets@0.1.0
├── file-icons@2.0.12
├── filesize@2.0.4
├── git-plus@7.2.0
├── haskell-grammar@0.4.0
├── highlight-selected@0.12.0
├── jump-to-line@0.2.0
├── language-babel@2.49.4
├── language-graphql@0.8.0
├── language-ini@1.17.0
├── language-lua@0.9.11
├── language-ocaml@1.1.2
├── language-swift@0.5.0
├── language-thrift@1.0.2
├── linter@1.11.21
├── linter-eslint@8.0.0
├── material-ui@1.0.9
├── merge-conflicts@1.4.4
├── nuclide@0.201.0
├── nuclide-format-js@0.0.36
├── oceanic-next@1.0.0
├── one-dark-vivid-syntax@1.7.1
├── pigments@0.38.0
├── pretty-json@1.6.1
├── project-manager@3.3.3
├── react-es6-snippets@0.3.0
├── set-syntax@0.3.2
[...]

PR #1017, opened 2017-02-06
Diagnostics: Add toggle to show full diagnostic traces

mwalkerwells:
@nmote In response to #998 cc @ColinCampbell 

It's common to need more information to debug flow errors. This change reduces the need to open terminal and run ``flow`` and doesn't clutter the interface.

Settings still has a "Show full diagnostic traces" option. It's used as a default value for when the "Diagnostics" view is constructed.

showinteraction

### DIFF
--- a/pkg/nuclide-diagnostics-ui/lib/main.js
+++ b/pkg/nuclide-diagnostics-ui/lib/main.js
@@ -135,9 +135,10 @@ class Activation {
         .switchMap(updater => (
           updater == null ? Observable.of([]) : updater.allMessageUpdates
         )),
-      featureConfig.observeAsStream('nuclide-diagnostics-ui.showDiagnosticTraces'),
+      // eslint-disable-next-line max-len
+      ((featureConfig.observeAsStream('nuclide-diagnostics-ui.showDiagnosticTraces'): any): Observable<boolean>),
       showTraces => {
-        featureConfig.set('nuclide-diagnostics-ui.showDiagnosticTraces', showTraces)
+        featureConfig.set('nuclide-diagnostics-ui.showDiagnosticTraces', showTraces);
       },
       disableLinter,
       observeLinterPackageEnabled(),
--- a/pkg/nuclide-diagnostics-ui/lib/main.js
+++ b/pkg/nuclide-diagnostics-ui/lib/main.js
@@ -140,8 +140,6 @@ class Activation {
           updater == null ? Observable.of([]) : updater.allMessageUpdates
         )),
       this._state.showTraces,
-      // https://github.com/gajus/eslint-plugin-flowtype/pull/131
-      // eslint-disable-next-line flowtype/space-after-type-colon
       showTraces => {
         if (this._state != null) {
           this._state.showTraces = showTraces;
@@ -150,8 +148,6 @@ class Activation {
       disableLinter,
       observeLinterPackageEnabled(),
       this._state.filterByActiveTextEditor,
-      // https://github.com/gajus/eslint-plugin-flowtype/pull/131
-      // eslint-disable-next-line flowtype/space-after-type-colon
       filterByActiveTextEditor => {
         if (this._state != null) {
           this._state.filterByActiveTextEditor = filterByActiveTextEditor;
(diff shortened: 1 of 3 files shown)


nmote:
Thanks! At first glance this looks pretty high quality, although I do have some comments. I'll take a closer look tomorrow.

pkg/nuclide-diagnostics-ui/lib/main.js
```diff
@@ -135,18 +139,24 @@ class Activation {
...
+      showTraces => {
+        if (this._state != null) {
+          this._state.showTraces = showTraces;
+        }
+      },
       disableLinter,
+      observeLinterPackageEnabled(),
+      this._state.filterByActiveTextEditor,
+      // https://github.com/gajus/eslint-plugin-flowtype/pull/131
+      // eslint-disable-next-line flowtype/space-after-type-colon
```
nmote: is this really still necessary?

mwalkerwells:
@nmote That's reasonable—I played around with that & realized I'd have to consider both being on screen at the same time (Settings & Diagnostics) and it got a bit more complicated to manage, so I went with this approach.

Thoughts on simplifying & having only in the Diagnostics panel?

If not, I'll try to figure out how to ensure neither get in an inconsistent state if both are on screen at the same time.

pkg/nuclide-diagnostics-ui/lib/main.js
mwalkerwells: Nope!

nmote:
I think that if you just subscribe to the setting (`featureConfig.observe`) and use that as the source of truth, and when the box is checked you call `featureConfig.set`, then you should have no problems with inconsistency. I'm not 100% sure that the settings panel will immediately react to outside changes to config options, but even if it doesn't I think this is a good way to go.

mwalkerwells:
Sorry for the thrash, but this works as expected: always correct state independent of being on-screen simultaneously.

One interesting side-effect is that it updates across all windows & is not isolated to a specific project.

pkg/nuclide-diagnostics-ui/lib/main.js
```diff
@@ -135,18 +135,19 @@ class Activation {
         .switchMap(updater => (
           updater == null ? Observable.of([]) : updater.allMessageUpdates
         )),
-      this._state.filterByActiveTextEditor,
-      // https://github.com/gajus/eslint-plugin-flowtype/pull/131
-      // eslint-disable-next-line flowtype/space-after-type-colon
-      (featureConfig.observeAsStream('nuclide-diagnostics-ui.showDiagnosticTraces'):
-          Observable<any>),
+      // eslint-disable-next-line max-len
```
nmote: Please don't do this. You can just split this line into multiple other lines.

mwalkerwells: Can you be more specific as to how you'd like this broken up on multiple lines (not sure of the best practices for splitting casting like this) or could you just edit it according to your preference before you accept (I'll take note for future pull requests)?

https://github.com/blog/1738-web-flow-editing-from-pull-requests

nmote: Looks like I can't change the PR, maybe since I don't have push access to your branch?

Anyway I'm not super picky how you format it. We have a rule on the team that you can't nitpick formatting at review time. If you care enough about anything you can write a lint rule for it. So, anything that passes lint is okay by me.

I'd suggest something like this:

```
((featureConfig.observeAsStream('nuclide-diagnostics-ui.showDiagnosticTraces'): any):
  Observable<boolean>),
```

mwalkerwells: 👍🏻

nmote:
Okay, I'm pulling it in for internal review so I can get another set of eyes on it.

matthewwithanm:
@mwalkerwells thanks for doing this! the "diagnostics" in the label is kinda redundant and "traces" should probably be "descriptions" since that's what they're called below.

also, how wide does the window have to be to accommodate this? we're getting a little tight with those two checkboxes. maybe we could shorten the other label too as an easy fix. ("Show only diagnostics for current file" is awkwardly worded anyway imo)

mwalkerwells:
The window can still get pretty narrow, especially since Atom changes the font size at a certain point. I'm happy to submit another diff with copy changes & will defer to your team's preference. Or if you already have consensus, I'll make the copy changes now

nmote:
@mwalkerwells It looks like @matthewwithanm is planning to just make the copy changes himself. This will land shortly.

mwalkerwells:
@nmote Sounds good!

PR #1021, opened 2017-02-09
[React Inspector] Bump react-devtools-core

gaearon:
Follow up to #1016.
Changes:

* Bumps `ws` dependency to 2.0 as [requested](https://github.com/facebook/nuclide/pull/1016#issuecomment-278046379).
* Adds support for the new DevTools injection mechanism that will very soon replace the current one in RN master (more details in https://github.com/facebook/react-devtools/issues/489).

Despite 2.0 bump this is backwards compatible for Nuclide (there were other changes it doesn't rely on).

Note that you can't test the change on Inspector with master of RN because the integration was recently temporarily disabled. However I tested it both with RN stable (works) and with pending changes I'm about to make to RN (also works).

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1838,9 +1838,16 @@
       "resolved": "https://registry.npmjs.org/react-addons-update/-/react-addons-update-15.3.1.tgz"
     },
     "react-devtools-core": {
-      "version": "1.0.6",
-      "from": "react-devtools-core@>=1.0.6 <2.0.0",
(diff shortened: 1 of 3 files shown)


matthewwithanm:
@hansonw @zertosh Are we okay to bump 2.0 now?

gaearon:
(Just bumped with bugfixes and stability improvements.)

zertosh:
Tracking this internally https://fburl.com/irscdq5c.

PR #1026, opened 2017-02-21
Upgrade to latest package-deps

steelbrain:
This PR

 - Upgrades to latest `atom-package-deps` version
 - Removes a config that is no longer necessary after the latest version of the package

You can see the new Confirmation Dialog in action in [`atom-package-deps` Screenshots](https://github.com/steelbrain/package-deps#screenshots)

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -275,7 +275,7 @@ export function activate() {
     const firstWindowId = remote.BrowserWindow.getAllWindows()[0].id;
     const currentWindowId = remote.getCurrentWindow().id;
     if (firstWindowId === currentWindowId) {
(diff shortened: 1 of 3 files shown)


hansonw:
@steelbrain Thanks for this - we'll actually upgrade to 4.5.0 if https://github.com/steelbrain/package-deps/pull/89 gets in! ;) A little hesitant to take on a bunch of extra dependencies that will be unused. Also:

1. `installRecommendedPackages` is reserved for internal use so we don't want to remove that
2. We'll want to leave `promptUser` to false, as we want to really ensure that our internal Nuclide users install these packages.

PR #1030, opened 2017-02-25
Added link to Atom documentation in Getting Started

zardra:
One more PR :)

### DIFF
--- a/docs/_docs/quick-start/getting-started.md
+++ b/docs/_docs/quick-start/getting-started.md
@@ -5,7 +5,10 @@ layout: docs
 permalink: /docs/quick-start/getting-started/
 ---
 
-This getting started guide walks you through the core features of Nuclide and aims to get you productive quickly.  See [Basics](/docs/editor/basics) for more information on using Nuclide's editing features.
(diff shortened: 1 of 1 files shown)


docs/_docs/quick-start/getting-started.md
```diff
@@ -5,7 +5,10 @@ layout: docs
 permalink: /docs/quick-start/getting-started/
 ---
 
-This getting started guide walks you through the core features of Nuclide and aims to get you productive quickly.  See [Basics](/docs/editor/basics) for more information on using Nuclide's editing features.
+This getting started guide walks you through the core features of Nuclide and aims to get you productive quickly.  
+
+If you're new to Atom, you can find more information about its features in the [Atom Flight Manual](http://flight-manual.atom.io/).  Then, to find out more about using Nuclide see [Basics](/docs/editor/basics) or any of the other Features, Plugins, or Language & Platform guides listed in the side menu for Nuclide-specific features.
```
JoelMarcey: I think we can be a little simpler here. And we may not always have a sidebar, possibly.

> Then, to find out more about using Nuclide, see [Basics](/docs/editor/basics) or any of the other guides available in the Nuclide documentation.  

What do you think?

PR #1062, opened 2017-03-21
Respect encoding on remote connection

mozisan:
Read/Write on remote connection works correctly only for files which encoding is `UTF-8`.
This PR fixes this.

### DIFF
--- a/pkg/nuclide-remote-connection/lib/RemoteFile.js
+++ b/pkg/nuclide-remote-connection/lib/RemoteFile.js
@@ -254,10 +254,11 @@ export class RemoteFile {
   }
 
   async read(flushCache?: boolean): Promise<string> {
-    const data = await this._getFileSystemService().readFile(this._localPath);
+    const encoding = this.getEncoding();
+    const options = encoding ? {encoding} : {};
+    const data = await this._getFileSystemService().readFile(this._localPath, options);
     const contents = data.toString();
     this._setDigest(contents);
-    // TODO: respect encoding
     return contents;
   }
 
@@ -267,7 +268,9 @@ export class RemoteFile {
 
   async write(text: string): Promise<void> {
     const previouslyExisted = await this.exists();
-    await this._getFileSystemService().writeFile(this._localPath, text);
+    const encoding = this.getEncoding();
+    const options = encoding ? {encoding} : {};
+    await this._getFileSystemService().writeFile(this._localPath, text, options);
     if (!previouslyExisted && this._subscriptionCount > 0) {
       this._subscribeToNativeChangeEvents();
     }
--- a/pkg/nuclide-server/lib/services/FileSystemService.js
+++ b/pkg/nuclide-server/lib/services/FileSystemService.js
@@ -255,11 +255,21 @@ export function unlink(path: string): Promise<void> {
  */
 export async function readFile(
   path: string,
-  options?: {flag?: string},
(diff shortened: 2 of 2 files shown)


pkg/nuclide-remote-connection/lib/RemoteFile.js
```diff
@@ -254,10 +254,11 @@ export class RemoteFile {
   }
 
   async read(flushCache?: boolean): Promise<string> {
-    const data = await this._getFileSystemService().readFile(this._localPath);
+    const encoding = this.getEncoding();
+    const options = encoding ? {encoding} : {};
```
hansonw: The Atom code always keeps this non-null/non-empty (and we've Flow-typed it as such), so I think that's fine to assume here as well. Let's just inline `options` as `{encoding}` below.

pkg/nuclide-server/lib/services/FileSystemService.js
```diff
@@ -255,12 +255,22 @@ export function unlink(path: string): Promise<void> {
...
 ): Promise<Buffer> {
   const stats = await fsPromise.stat(path);
   if (stats.size > READFILE_SIZE_LIMIT) {
     throw new Error(`File is too large (${stats.size} bytes)`);
   }
+  if (options && options.encoding) {
+    const encoding = options.encoding;
+    delete options.encoding;
+    if (encoding !== 'utf8') {
+      const iconv = require('iconv-lite');
```
hansonw: Nuclide uses a transform to lazy-require all imports, so just import this at the top of the file in the usual way.

pkg/nuclide-remote-connection/lib/RemoteFile.js
```diff
@@ -254,10 +254,11 @@ export class RemoteFile {
...
+    const data = await this._getFileSystemService().readFile(this._localPath, options);
     const contents = data.toString();
```
hansonw: Something about this feels off. Does this work? Doesn't this implicitly encode the buffer as a UTF-8 string? Or does this work because the `readFile` call implicitly decodes into a UTF-8 string?

Is that guaranteed to work? IMO it would be clearer and more preferable to keep the `readFile` call encoding-agnostic, let it keep returning a raw Buffer, and do the decoding here (as with Atom's implementation).

pkg/nuclide-server/lib/services/FileSystemService.js
```diff
@@ -314,7 +324,14 @@ export async function writeFile(path: string, data: string,
   let complete = false;
   const tempFilePath = await fsPromise.tempfile('nuclide');
   try {
-    await fsPromise.writeFile(tempFilePath, data, options);
+    if (options && options.encoding && options.encoding !== 'utf8') {
+      const iconv = require('iconv-lite');
```
hansonw: ditto

hansonw (changes requested):
Sorry about the delayed response! some comments inline.

Another note: we don't have `iconv-lite` as an explicit dependency (though it is included transitively). You'll want to add this to our `package.json` to make this explicit.

Probably use `~0.4.13` to avoid duplication with the existing version (which comes from `node-fetch` -> `encoding`)

hansonw:
If you're using Atom 1.19 the new remote file implementation should work with encodings!
https://github.com/facebook/nuclide/blob/master/pkg/nuclide-remote-connection/lib/RemoteFile.js#L355

Issue #1098, opened 2017-04-13
Diagnostics view looks wrong when "Current file only" checked


jazmon:
### Issue and Steps to Reproduce
The issue emerged with https://github.com/facebook/nuclide/commit/e9f515aaff3e19bf48424e3307b0a31db3f2dba5.  The commit seems to change line number positioning to the right, but does it naïvely by assuming it's the same position every time. This is not the case when "Current file only" is checked.

#### Reproducing:
1. Open a project w/ flow or eslint enabled
2. Open file with errors
3. In diagnostics panel check "Current file only"

### Expected Behavior
Description column's text should be aligned to the left.
![Expected](https://puu.sh/viM8i/e2a45de902.png)

### Actual Behavior
Description column's text is aligned to the right.
![Actual](https://puu.sh/viM8E/38ca6ec9ab.png)
### Versions
* Atom: 1.16.0
* Nuclide: 0.217.0
* Client OS: macOS 10.12.4
* Server OS (optional):

### Additional Details
* Installed packages (`apm ls --installed`):
```├── api-docs@0.0.6
├── atom-beautify@0.29.18
├── autoclose-html@0.23.0
├── autocomplete-emojis@2.5.0
├── code-peek@1.4.18
├── color-picker@2.2.5
├── docblockr@0.9.4
├── double-tag@0.8.0
├── editorconfig@2.2.2
├── elm-format@2.1.0
├── elmjutsu@5.5.1
├── emmet@2.4.3
├── file-icons@2.1.2
├── haskell-grammar@0.4.0
├── highlight-selected@0.13.1
├── html-to-elm@0.2.0
├── language-babel@2.57.2
├── language-elm@1.5.0
├── language-groovy@0.6.0
├── language-ini@1.19.0
├── language-ocaml@1.8.0
├── language-scala@1.1.8
├── language-swift@0.5.0
├── language-thrift@1.0.2
├── lebab@0.10.0
├── linter@2.1.2
├── linter-elm-make@0.23.2
├── linter-eslint@8.1.5
├── linter-stylelint@3.10.0
├── linter-tidy@2.3.1
├── merge-conflicts@1.4.4
├── minimap@4.27.1
├── minimap-bookmarks@0.4.2
├── nuclide@0.217.0
├── pigments@0.39.1
├── prettier-atom@0.26.0
├── project-manager@3.3.4
├── sort-lines@0.14.0
├── stylefmt@4.1.0
├── sync-settings@0.8.1
[...]

PR #1102, opened 2017-04-16
Fix diagnostics view aligning description column right in some cases

jazmon:
Related issue: https://github.com/facebook/nuclide/issues/1098.

Basically this PR:
* Adds an optional field that whether a certain column should align text right
* Changed the css rule for line-number to use the optional field, instead of assuming it's always the fourth column.


matthewwithanm:
Wow, sorry this slipped through the cracks @jazmon. Thanks!

jazmon:
Heh, no worries! Glad it's being merged :)

PR #1123, opened 2017-05-02
Add a 'kind' property to allow the UI to be easily themed by kind

damieng:
This PR adds a kind property that can be set by the underling provider that bubbles up to the ui as a `kind-${outline.kind}` up on the `<li>` for the item, e.g.

image

Does not interfere with the current icon property/visualization (although could be overridden by a theme)

### DIFF
--- a/pkg/nuclide-outline-view/lib/OutlineView.js
+++ b/pkg/nuclide-outline-view/lib/OutlineView.js
@@ -173,7 +173,7 @@ class OutlineTree extends React.PureComponent {
       }
     };
 
-    const classNames = [ 'list-nested-item' ];
+    const classNames = ['list-nested-item'];
     if (outline.kind) {
       classNames.push(`kind-${outline.kind}`);
     }
--- a/pkg/nuclide-outline-view/lib/OutlineView.js
+++ b/pkg/nuclide-outline-view/lib/OutlineView.js
@@ -173,7 +173,11 @@ class OutlineTree extends React.PureComponent {
       }
     };
 
-    const classes = classnames('list-nested-item', {
+    const classNames = [ 'list-nested-item' ];
+    if (outline.kind) {
+      classNames.push(`kind-${outline.kind}`);
+    }
+    const classes = classnames(classNames, {
       selected: outline.highlighted,
     });
     return (
--- a/pkg/nuclide-outline-view/lib/main.js
+++ b/pkg/nuclide-outline-view/lib/main.js
@@ -41,27 +41,6 @@ export type OutlineTreeForUi = {
   highlighted: boolean,
 };
 
-// Kind of outline tree - matches the names from the Language Server Protocol v2.
-export type OutlineTreeKind =
-  | 'file'
-  | 'module'
-  | 'namespace'
-  | 'package'
-  | 'class'
-  | 'method'
-  | 'property'
-  | 'field'
-  | 'constructor'
-  | 'enum'
-  | 'interface'
-  | 'function'
-  | 'variable'
-  | 'constant'
-  | 'string'
-  | 'number'
-  | 'boolean'
-  | 'array';
-
 /**
  * Includes additional information that is useful to the UI, but redundant or nonsensical for
  * providers to include in their responses.
--- a/pkg/nuclide-outline-view/lib/main.js
+++ b/pkg/nuclide-outline-view/lib/main.js
@@ -27,11 +27,11 @@ import {createOutlines} from './createOutlines';
 import {Observable} from 'rxjs';
 
 import type {TokenizedText} from '../../commons-node/tokenizedText-rpc-types';
-import type {Outline} from './rpc-types';
+import type {Outline, OutlineTreeKind} from './rpc-types';
 
 export type OutlineTreeForUi = {
   icon?: string, // from atom$Octicon, but we use string for convenience of remoting
-  kind?: string, // kind you can pass to the UI for theming, from OutlineTreeKind
+  kind?: OutlineTreeKind, // kind you can pass to the UI for theming
   plainText?: string,
   tokenizedText?: TokenizedText,
 
--- a/pkg/nuclide-outline-view/lib/main.js
+++ b/pkg/nuclide-outline-view/lib/main.js
@@ -43,9 +43,24 @@ export type OutlineTreeForUi = {
 
 // Kind of outline tree - matches the names from the Language Server Protocol v2.
 export type OutlineTreeKind =
-  'file' | 'module' | 'namespace' | 'package' | 'class' | 'method' | 'property' | 'field' |
-  'constructor' | 'enum' | 'interface' | 'function' | 'variable' | 'constant' | 'string' |
-  'number' | 'boolean' | 'array';
+  | 'file'
+  | 'module'
+  | 'namespace'
+  | 'package'
+  | 'class'
+  | 'method'
+  | 'property'
+  | 'field'
+  | 'constructor'
+  | 'enum'
+  | 'interface'
+  | 'function'
+  | 'variable'
+  | 'constant'
+  | 'string'
+  | 'number'
+  | 'boolean'
+  | 'array';
 
 /**
  * Includes additional information that is useful to the UI, but redundant or nonsensical for
--- a/pkg/nuclide-outline-view/lib/main.js
+++ b/pkg/nuclide-outline-view/lib/main.js
@@ -31,7 +31,7 @@ import type {Outline} from './rpc-types';
 
 export type OutlineTreeForUi = {
   icon?: string, // from atom$Octicon, but we use string for convenience of remoting
-  kind?: string, // a kind you can pass to the UI for theming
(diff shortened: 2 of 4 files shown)


pkg/nuclide-outline-view/lib/OutlineView.js
```diff
@@ -173,7 +173,11 @@ class OutlineTree extends React.PureComponent {
       }
     };
 
-    const classes = classnames('list-nested-item', {
+    const classNames = [ 'list-nested-item' ];
```
hansonw: We've taken to using `prettier` to format our code so you gotta remove these spaces

jgebhardt:
What is the set of accepted `kind` values? I think standardizing, or at least publicly documenting, the set of possible values will be important for making the feature work reliably across multiple themes and providers.

Since there's no validation that the "kind name" is actually a valid css className (e.g. that it doesn't contain whitespace), I'd suggest simplifying this API to allow passing a classname rather than `kind` in the outline- this will at least make the className semantics clear. The above point about standardization would still apply though.

damieng:
We should probably just use the same 'kind' values as the language server protocol. We should either accept only lowercase or accept title case and lower-case the css attribute.

```javascript
export namespace SymbolKind {
	export const File = 1;
	export const Module = 2;
	export const Namespace = 3;
	export const Package = 4;
	export const Class = 5;
	export const Method = 6;
	export const Property = 7;
	export const Field = 8;
	export const Constructor = 9;
	export const Enum = 10;
	export const Interface = 11;
	export const Function = 12;
	export const Variable = 13;
	export const Constant = 14;
	export const String = 15;
	export const Number = 16;
	export const Boolean = 17;
	export const Array = 18;
}
```

ljw1004:
Damien I'm not familiar with this codebase so I'm having a hard time following through. Would you mind writing out a concrete example of the HTML that will be produced? And then what would the theme do to skin it? (Also, is a theme able to introduce new icons? PNG or SVG? do themes do this already?)

Also, we're going to be asking LSP to add a new "kind" for Trait. In general, if LSP adds new kinds, what will existing icon themes be like?


Incidentally, several people have approached me to say "I want the same icons as IntelliJ" and others to say "I want the same icons as Eclipse, not IntelliJ". So there's a definite wish to theme the icon-sets.

They also want, like Eclipse+IntelliJ, to use icon representation of more information - public/private/protected for sure. In VisualStudio those are represented by small mini-glyphs superimposed in the bottom right corner. Anyway, those thoughts are for the future, unrelated to this current PR.

damieng:
The extra code here just adds a single CSS attribute to the LI - it is shown highlighted in orange on the screenshot attached to the PR.  

A theme author could then do something like this

```css
.nuclide-outline-view span.icon {
  display: none;
}

.nuclide-outline-view .kind-property div {
  background-repeat: no-repeat;
  background-position:  0 50%;
  background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAABGdBTUEAAK/INwWK6QAAABl0RVh0U29mdHdhcmUAQWRvYmUgSW1hZ2VSZWFkeXHJZTwAAAFGSURBVDjLY/j//z8DJZhhcBtwa5ou062Zhp0PV7udvjVNO5lkA25M1Cx4czL//+/PJ/8/3xX+9myrqgfRBtycppf26mDs398fFv//cKXh/5vjdf/vLg74ui1PMZIoA+4utF7/893K/z+e1f3/8WLW/w8X6/5/e9Tw/+YM+w9EGXBrpknbl7ud/789yPr//VHx/2/34/9/uxX6/2y9xWe8BtycqmvycJXbqac7Au59uhn3/9ttv//f70b//3oDqLlB7/uiCAXcXrg1wyDm7jyrc98ezfz/9U7w/2833f6fmWD+5Xyz6YtLLWYvlscpBOMNxIerPc7/eLbw/6dLFv+/XXP/f3aK9deZ0cpJO4KleA/EyfISjMbH6zxO/nzZ9v/rTd//ZyZZfZ6bqBJLUkI62WvSenW+68vTvVYPlqarhg2DvEAMBgD1ZuZTUbWrEgAAAABJRU5ErkJggg==);
}

.nuclide-outline-view .kind-method div {
  background-repeat: no-repeat;
  background-position:  0 50%;
  background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAABGdBTUEAAK/INwWK6QAAABl0RVh0U29mdHdhcmUAQWRvYmUgSW1hZ2VSZWFkeXHJZTwAAAEsSURBVDjLY/j//z8DJZhhmBuQvdj8a+p8w//xc3U5yTIgeb7h18bNUf/DZ2j8958qz0nQgPyltv9zF5v/TV9o/Ddxrv7fmvWh/1ednvi/ZLX/f9d+8b+23YI8eA0AOvn/hnPT/q89OwWsccXpCf8n7Cn5v/B41/+MpW7/TdvZ/+o2M/LjNADoZLDmvl35/zt3ZP9v3Zb2v2Fz4v+mren/Zxxq/h+zwOa/aj3DH5wGRM/W/L/y1IT/S0/0/l94rOv/vKMdQEOy/k8/1PQ/banHf8VahlfSlQwGOA0InK74z3Oi9D/nftF/1t38/+LmW/2fdrDhf9Jit//ytQzPJSsZtEiKBe1mxq/xC53/y1czPAFqVic5GoFO/ipXzfxftJJBkeyUKFzOwDm48wIAh5XH+g7drOwAAAAASUVORK5CYII=);
}

.nuclide-outline-view span {
  padding-left: 5px
}```

In the case shown there the icons are base64 encoded png's but they could be resources in the theme pack, jpg, gifs, svg etc.

That theme with my PR produces this;

![image](https://cloud.githubusercontent.com/assets/118951/25715219/9fbdbd1c-30af-11e7-8b95-28c21e2d2777.png)

We could provide a bunch of default icons - I can work with our designer to include a default set of icons. As new ones are added to the spec we'd add icons for those. A theme author could specify their own default using CSS and if they did non then they would get our default icon for the new ones until they update their theme.

Once LSP supports visibility, traits etc. we can just add them like this as "trait-interface" "access-private" etc. and further overlay icons etc.

**Comment by ljw1004** [comment]
Okay, so I guess if we did add a new kind called "trait", and a theme didn't yet provide an icon for that theme, then it would fall back on whatever in-the-box CSS style was present for "kind-trait", which would look out-of-place but at least would look like something. That's fine.

**Comment by damieng** [comment]
Should I update the PR to make the changes to flow-type the list of current possible values for kind then?

**Comment by damieng** [comment]
Also do we want to call it something other than 'kind' on the css and/or outline view object given that there is already a 'kind' at the outline level for empty/loading etc.

**Comment by hansonw** [comment]
Yeah sounds like we'd be happy to take this with the flow-typed list. Using the LSP set seems good (and then we can even start doing the kind -> icon mapping in one place so we don't have to do it for every provider!)

I think:

```
export type OutlineTreeKind =
  'file' | 'module' | 'namespace' | ....
```

should work. "`kind`" seems like a totally reasonable name to me.

Note that our JS formatter (prettier) is pretty particular about how this is formatted; run `eslint` and it should be able to check your formatting.

File: pkg/nuclide-outline-view/lib/main.js [hunk 2]
```diff
@@ -40,6 +41,27 @@ export type OutlineTreeForUi = {
   highlighted: boolean,
 };
 
+// Kind of outline tree - matches the names from the Language Server Protocol v2.
+export type OutlineTreeKind =
```
hansonw: It looks like this should be moved to `rpc-types.js` so you can actually use it from both `OutlineTreeForUi` and `OutlineTree`. Or is there something blocking that?

pkg/nuclide-outline-view/lib/rpc-types.js
```diff
@@ -13,6 +13,7 @@ import type {TokenizedText} from '../../commons-node/tokenizedText-rpc-types';
 
 export type OutlineTree = {
   icon?: string, // from atom$Octicon (that type's not allowed over rpc so we use string)
+  kind?: string, // kind you can pass to the UI for theming, from OutlineTreeKind
```
hansonw: Can you actually make this `kind?: OutlineTreeKind` now? (Same above).

damieng: I'm not sure what restrictions your RPC stuff has. The one above is supposed to be from the list of atom$Octicons but is also defined as string as "that type's not allowed over rpc" - do you have any docs about the RPC process or steps I can test this with?

hansonw: Ohh yeah.. unfortunately we don't have great docs about this :( `atom$Octicon` is not supported because that's a type from `flow-libs`, which the RPC framework doesn't know about. Rest assured that simple literal unions like this one should work fine.

There's an option in Nuclide settings called "Use local RPC" that is sufficient for testing this (just make sure Flow outlines still work, for example).

PR #1134, opened 2017-05-12
Use Markdown markup to denote Remote Development Prerequisites list

vtambourine:
Markdown-style link are not working inside HTML tags. You can see the broken links on [Remote Development](https://nuclide.io/docs/features/remote/#prerequisites) page. Here I tried to fix it by using regular plain list syntax instead of `ul` tags.

Correct me if I wrong, but the only reason to use HTML here is to have `<li id="node"></li>` item, which, probably, have to be `class="node"`, but still not used anywhere.


hansonw:
Good catch. Thanks!

Issue #1185, opened 2017-06-10
index.html missing html code and looks incomplete


KrishMunot:
### Issue and Steps to Reproduce
I'm curious to understand the purpose of this directory - https://github.com/facebook/nuclide/tree/master/docs/docs

### Expected Behavior
The index.html in that directory was supposed to have HTML code, it looks like it isn't complete perhaps, does it need to be developed further?

### Actual Behavior
It looks like a templated text in a HTML file

### Versions
* Atom:
* Nuclide:
* Client OS:
* Server OS (optional):

### Additional Details
* Installed packages (`apm ls --installed`):

captbaritone:
Question answered in #1186

PR #1186, opened 2017-06-10
Make the file md readable

KrishMunot:
Realized that this file was supposed to a be a markdown and not a html looking at the code / content in the file.
Might potentially fix this issue that I filed - https://github.com/facebook/nuclide/issues/1185


captbaritone:
This file is not really html or markdown, it just contains header metadata which is consumed by Jekyll. Jekyll is equally happy to read HTML or Markdown. I guess the only real difference is that GitHub does a better job of formatting Jekyll metadata in markdown files.

PR #1188, opened 2017-06-10
fix image paths

KrishMunot:


KrishMunot:
What's wrong with these? The image paths were incorrect earlier

ebluestein:
The paths are currently correct (see https://nuclide.io/static/images/blog/2017-02-27/poll.png)

PR #1189, opened 2017-06-10
fixing image paths

KrishMunot:


ebluestein:
These paths are incorrect.

PR #1247, opened 2017-07-26
Add openExternal setting

danielbayley:
Allow users to set globally whether links are opened externally (the default) or indicate that links should be opened within Atom.

### DIFF
--- a/modules/atom-ide-ui/pkg/hyperclick/README.md
+++ b/modules/atom-ide-ui/pkg/hyperclick/README.md
@@ -57,7 +57,7 @@ export function getProvider() {
         // The range(s) to underline as a visual cue for clicking.
         range,
         // The function to call when the underlined text is clicked.
-        callback() {},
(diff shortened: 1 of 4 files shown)


shushz:
But hyperclick providers are not limited to just opening URLs. In fact, the most used scenario is navigating the source code.

danielbayley:
@shushz Sure. This doesn't actually enforce anything… it just provides a way for plugins that _do_ handle URLs to respect a single setting of whether to open them externally or within Atom. (There are a [bunch](https://atom.io/packages/search?q=browser) of browser plugins…)

shushz:
Sorry, I'm not convinced. Not every clickable link in the UI comes from the hyperclick package, so this still won't be a great place for such setting. What if you have a link in the diagnostics panel, or console, or terminal. So, if anything such setting should be Atom-global.

And for the providers that don't handle URIs this would make a poor API.

PR #1291, opened 2017-09-07
[nuclide][code search] fix last bugs

a20012251:
On windows, it's necessary to have the 'FILE' type in the search results, otherwise the links don't open
Also, show the relative paths with a beginning . and with / instead of \

image


a20012251:
already commited internally

PR #1306, opened 2017-09-19
Open new files in a pending state in goToLocation

mehcode:
Tracked the source of this behaviour down when using `ctrl-click` in Atom IDE. I feel that pending tabs are the right UX when I'm navigating around a codebase. Remember that pending tabs can be turned off globally in Atom if a user dislikes them.


hansonw:
We can just keep this PR on facebook-atom/atom-ide-ui: I'll sync it back up afterward :)

PR #1473, opened 2018-02-27
Updated documentation for the split pane keybinding

vsrivatsan89:
The keybinding for splitting pane is defined as Cmd-K-<arrow> but I believe it is Cmd-K,<arrow>


hansonw:
Seems reasonable. Thank you!

PR #1486, opened 2018-03-23
Add Option to change font for nuclide-terminal

marudor:
I like to use the internal Terminal once in a while. But as xterm.js renders the terminal in a canvas I have to set the font in xterm.js and can't just change the css.
Sadly nuclide didn't have an option to change the font.
Now it does.

### DIFF
--- a/pkg/nuclide-terminal/lib/terminal-view.js
+++ b/pkg/nuclide-terminal/lib/terminal-view.js
@@ -54,6 +54,8 @@ const SCROLLBACK_CONFIG = 'nuclide-terminal.scrollback';
 const CURSOR_STYLE_CONFIG = 'nuclide-terminal.cursorStyle';
 const CURSOR_BLINK_CONFIG = 'nuclide-terminal.cursorBlink';
 const FONT_FAMILY_CONFIG = 'nuclide-terminal.fontFamily';
+const FONT_SIZE_CONFIG = 'nuclide-terminal.fontSize';
+const LINE_HEIGHT_CONFIG = 'nuclide-terminal.lineHeight';
 const DOCUMENTATION_MESSAGE_CONFIG = 'nuclide-terminal.documentationMessage';
 const ADD_ESCAPE_COMMAND = 'nuclide-terminal:add-escape-prefix';
 const TMUX_CONTROLCONTROL_PREFIX = '\x1BP1000p';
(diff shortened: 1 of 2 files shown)


marudor:
Eh... Test just "died unexpectedly"
No Idea what happened there.

pelmers:
thanks for the contribution! This feature has been requested internally often as well.

re: 4gb limit, that's been happening often lately and generally I just press the re-run button.

pkg/nuclide-terminal/lib/terminal-view.js
```diff
@@ -53,6 +53,7 @@ const PRESERVED_COMMANDS_CONFIG = 'nuclide-terminal.preservedCommands';
 const SCROLLBACK_CONFIG = 'nuclide-terminal.scrollback';
 const CURSOR_STYLE_CONFIG = 'nuclide-terminal.cursorStyle';
 const CURSOR_BLINK_CONFIG = 'nuclide-terminal.cursorBlink';
+const FONT_FAMILY_CONFIG = 'nuclide-terminal.fontFamily';
```
pelmers: while we're at it, would you also be willing add terminal-specific font size and line height settings?

pkg/nuclide-terminal/lib/terminal-view.js
```diff
@@ -202,6 +204,10 @@ export class TerminalView implements PtyClient {
         .observeAsStream(SCROLLBACK_CONFIG)
         .skip(1)
         .subscribe(scrollback => terminal.setOption('scrollback', scrollback)),
+      featureConfig
+        .observeAsStream(FONT_FAMILY_CONFIG)
+        .skip(1)
+        .subscribe(fontFamily => terminal.setOption('fontFamily', fontFamily)),
```
pelmers: instead of adding a new call to setOption for fontFamily (which will conflict with the existing one that sync the Atom text editor setting), could you instead rename existing 'editor.fontFamily' to be FONT_FAMILY_CONFIG (line 136)?

Then change syncAtomStyleItem (~line 400) to match the new terminal group (ideally you could also add options for the fontSize and lineHeight).

marudor: I'm not exactly sure what syncAtomStyle does and why it is there - I've changed it to use the new Option but instead of using the existing I just set the options directly.
Using syncAtomStyle made it slow and sluggish on several occasions.
No Idea why exactly - Not that familiar with the nuclide or atom code.

Doing it like this works though.

Issue #1341, opened 2017-10-13
Remote development => utils - ERROR - Unexpected error: <type 'exceptions.AttributeError'>


fanch317:
### Issue and Steps to Reproduce
When I try to add a Remote Project Folder, it returns this error:
Unexpected error occurred: The log file can be found at /tmp/nuclide-dev-logs/server-start/nuclide-2017-10-13T21:54:04.nohup.out. We can see many entries **utils - ERROR - Unexpected error: <type 'exceptions.AttributeError'>** on server log.

### Expected Behavior
Starts and works remotely over nuclide server.

### Actual Behavior
I start a remote connection from client Atom, and see many error log "_Unexpected error_" (101 entries on 10s).

This is log with verbose mode : _nuclide-start-server -v_

nuclide.log
> 2017-10-13 21:49:25.455 INFO (pid:29103) nuclide-server - NuclideServer started on port 9090.
> 2017-10-13 21:49:25.457 INFO (pid:29103) nuclide-server - Using node v8.7.0.
> 2017-10-13 21:49:25.457 INFO (pid:29103) nuclide-server - Server ready time: 236ms

server-start/nuclide.log
> 2017-10-13 21:49:25,081 - root - INFO - Invoked nuclide_server_manager...
> 2017-10-13 21:49:25,081 - NuclideServerManager - INFO - NuclideServerManager was created with these options: {'certs_dir': None, 'json_output_file': None, 'timeout': 10, 'dump_core': True, 'insecure': False, 'inspect': False, 'quiet': False, 'debug': False, 'command': 'start', 'workspace': None, 'common_name': None, 'port': None, 'verbose': True}
> 2017-10-13 21:49:25,081 - NuclideServerManager - INFO - Starting to clean up old Nuclide processes/files.
> 2017-10-13 21:49:25,099 - NuclideServerManager - INFO - Cleaning up old certificate files...
> 2017-10-13 21:49:25,100 - NuclideServerManager - INFO - Finished cleaning up old Nuclide processes/files.
> 2017-10-13 21:49:25,100 - root - WARNING - Failed to clean up old core dumps ([Errno 2] No such file or directory: '/var/tmp/cores')
> 2017-10-13 21:49:25,100 - NuclideServerManager - INFO - Trying to determine the port to use for Nuclide server...
> 2017-10-13 21:49:25,117 - NuclideServerManager - INFO - Found an open port: 9090.
[...]

PR #1483, opened 2018-03-21
Fixing regex for get openssh common_name #1341

Empornium:

PR #1487, opened 2018-03-25
Fixing regex for get openssh common_name #1341

fanch317:
#1341


captbaritone:
@hansonw

hansonw:
Imported this already - see #1483. Thanks for coming up with the fix! :)

PR #1492, opened 2018-04-03
fix homebrew openssl path

jrunge:
Not sure what version of homebrew placed openssl in /opt/homebrew/bin, but the current installer places it in /usr/local/opt/openssl/bin.

I'm assuming openssl has not been linked in homebrew (made the default), or the path addition would be unnecessary.

I was unable to run nuclide-start-server on High Sierra (10.13.4) until I made this change. This seems to be because LibreSSL does not support environment variables in the included openssl.cnf.

### DIFF
--- a/modules/big-dig/src/server/certificates.js
+++ b/modules/big-dig/src/server/certificates.js
@@ -189,7 +189,7 @@ function generateEnvironmentForOpenSSLCalls(serverCommonName: string): Object {
   if (process.platform === 'darwin') {
     // High Sierra comes with LibreSSL by default, which is not supported.
     // Often, OpenSSL may be installed by Homebrew.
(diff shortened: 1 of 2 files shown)


wanderley:
@hansonw I think it should be safer to add both `/opt/homebrew/bin` and `/usr/local/opt/openssl/bin` in the `PATH` (I do have both directories in my computer).

hansonw:
Yeah that seems correct (I might've made some assumptions about Facebook machines). Thanks!

jrunge:
No worries, thanks for looking at this so quickly. Totally makes sense to add both paths, in case the old one was being used in some cases. I'm just looking forward to being able to upgrade without breaking my setup. Remote projects is by far the best remote editing solution I've found. Cheers!

PR #1495, opened 2018-04-05
Update nuclide_server_manager.py

jzheng636:
For machines with ipv6 disabled (yes it is very strange, but...), the current script would throw. This change adds protection against this situation, so that it can still work correctly.


wanderley:
cc @hansonw

jzheng636:
Thanks @hansonw , sorry I did not check email for a while.  Made change as you suggested.

Issue #1429, opened 2018-01-03
I cannot get autocompletion when using .es6 file extension


emanuelbsilva:
### Issue and Steps to Reproduce
When using .es6 file extension autocompletion does not work. The moment i change the file extension of the file to .js it starts working.

I already tried to use the config.cson atom file to add a customFileType to "source.js" but it still doesn't work.

### Versions
* Atom: 1.23.2
* Nuclide: 0.273.0
* Client OS: 10.12.6 (16G29)
* Server OS (optional):

### Additional Details
* Installed packages (`apm ls --installed`):

emanuelbsilva:
I did more research on the issue and found out the origin of it. It appears that the JSAutoImports autocomplete provider is not yielding any result for some reasona nd that makes Atom not being able to present any autocomplete results because it's still waiting for JSAutoImports i'm assuming.

Taking a close look at the `https://github.com/facebook/nuclide/blob/master/pkg/nuclide-js-imports-client/lib/main.js` file i found out this code:
```
const lspService = await jsService.initializeLsp(
    ['.flowconfig'],
    ['.js'],
    (featureConfig.get('nuclide-js-imports-client.logLevel'): any),
    fileNotifier,
    host,
    getAutoImportSettings(),
  );
  return lspService || new NullLanguageService();
```

When i added `'.es6'` to the `['.js']` array autocomplete started working.

I think the best approach to solve the issue is to use whatever extension Atom's `config.cson` has registered for `source.js`.

PR #1544, opened 2018-06-01
Fix .jsx extension for #1429

PinkaminaDianePie:
Quick fix for the bug, which prevents using `nuclide-js-imports-client` with `.jsx` files extension. #1429


hansonw:
Makes sense! Re the original issue: do you mind also adding this to `nuclide-flow`? Flow is actually the main provider of autocomplete suggestions for JS files in Nuclide.

https://github.com/facebook/nuclide/blob/master/pkg/nuclide-flow/lib/main.js#L147

PinkaminaDianePie:
@hansonw thanks for mention this, done :)

hansonw:
Perfect. Thanks for your contribution!

PR #1606, opened 2018-08-31
WIP: Add support for Rust

Xanewok:
This sets up preliminary support for Rust in Buck.
It uses the [Rust Language Server (RLS)](https://github.com/rust-lang-nursery/rls) LSP server to provide necessary IDE features.

Since RLS and Cargo (the Rust package manager) don't support custom build systems out of box yet, this uses a hacky workaround for the moment. The user has to explicitly build a Rust target in Buck using the Buck toolbar - this in turn executes the `#save-analysis` flavor build and passes location of the resulting JSON save-analysis (index) files to the RLS. For this, the files should be saved first, before executing the build.

Until https://github.com/rust-lang/rust/pull/53870 lands and the updated RLS will ship with a Rust toolchain, it has to be built manually using recent Rust nightly compiler and to have its executable [pointed at](https://github.com/facebook/nuclide/commit/828e5e35e3a389b2e3ca62b619de1c9f88511dff#diff-08dcd94d8ccb34343d6c87bc397b1b59R26) by the user.

Currently the workflow is fragile - if the users starts editing the files without the explicit build first, the RLS can crash because how much it's currently tied to Cargo. This should be fixed when https://github.com/rust-lang-nursery/rls/issues/1026 is implemented (at least stop it from crashing).

cc @jsgf @mostafaeweda @sunshowers


pkg/nuclide-rust/lib/BuckIntegration.js
```diff
@@ -0,0 +1,87 @@
...
+const logger = getLogger('nuclide-rust');
+
+export async function updateRlsBuildForTask(
+  task: TaskInfo,
+  service: AtomLanguageService<LanguageService>,
+) {
+  const buildTarget = normalizeNameForBuckQuery(task.buildTarget);
+
+  // TODO: Filter by known Rust build targets
+  const files = await getRustInputs(task.buckRoot, buildTarget);
```
Xanewok: This still requires explicit filtering for `rust_*` target kinds.

Xanewok:
@jsgf I added a disclaimer (https://github.com/facebook/nuclide/pull/1606/commits/b948e1f1875b0e54d419891f02c6abab55a5c3d6) that warns the user to run the Buck build in a Rust project first. Is that desirable/okay for now?

jsgf:
@Xanewok I think its OK as a placeholder. I hope we can get something more elegant in the longer term though.

Xanewok:
> I hope we can get something more elegant in the longer term though.

Definitely, I am conviced this will not be an issue anymore once the RLS natively supports custom build systems.

Update: https://github.com/rust-lang/rust/pull/53870 has been merged and the newest Rust nightly (2018-09-04) contains the RLS binary supporting external simple build systems (needed for this) out of box, as well as external Rustfmt.

pkg/nuclide-rust/lib/BuckIntegration.js
mostafaeweda: please do this filtering to avoid calling the expensive `buck query` unless we're sure this is a rust target & this call will have value.

pkg/nuclide-rust/lib/BuckUtils.js
```diff
@@ -0,0 +1,57 @@
...
+  buildTarget: BuildTarget,
+): Promise<Array<string>> {
+  return BuckService.query(
+    buckRoot,
+    `filter('.*\\.rs$', inputs(kind('^rust_.*', ${buildTarget})))`,
+    [],
+  );
+}
+
+export function getSaveAnalysisTargets(
```
mostafaeweda: change to use async / await :
```js
export async function getSaveAnalysisTargets(
...
  const deps = await BuckService.query(buckRoot, query, []);
  return deps.map(dep => dep + '#save-analysis');
}
```

pkg/nuclide-rust/lib/BuckIntegration.js
```diff
@@ -0,0 +1,96 @@
...
+  // Instead, we build now, copy paths to resulting .json analysis artifacts to
+  // a temp file and just use `cat $TMPFILE` as a dummy build command.
+  const analysisTargets = await getSaveAnalysisTargets(
+    task.buckRoot,
+    buildTarget,
+  );
+  logger.debug(`analysisTargets: ${analysisTargets.join('\n')}`);
+  const artifacts: Array<string> = [];
+
+  const buildReport = await BuckService.build(task.buckRoot, analysisTargets);
```
mostafaeweda: how long does this typically take for large rust targets? do we need to report its progress?

pkg/nuclide-rust/lib/BuckIntegration.js
```diff
@@ -0,0 +1,96 @@
...
+  const files = await getRustInputs(task.buckRoot, buildTarget);
+  // Probably not a Rust build target, ignore
+  if (files.length === 0) {
+    return;
+  }
+  // We need only to pick a representative file to get a related lang service
+  const fileUri = task.buckRoot + '/' + files[0];
+  logger.debug(`fileUri: ${fileUri}`);
+
+  const langService = await service.getLanguageServiceForUri(fileUri);
```
mostafaeweda: There's an invalid assumption here that the multi-project language service will only have one LSP server that we need to update its build command (while in reality there can be multiple buck roots, each having a different LSP server -- according to the project files config below).

Do you need to `getAllLanguageServices` or `observeLanguageServices`?
Can you test with multiple buck roots to see if things work properly?

pkg/nuclide-rust/lib/main.js
```diff
@@ -0,0 +1,58 @@
...
+
+import {updateRlsBuildForTask} from './BuckIntegration';
+
+const DISCLAIMER = `[nuclide-rust] Support for Buck-managed Rust
+projects is currently experimental. For it to work correctly, please build
+the target you plan on working using Buck toolbar.`;
+
+class Activation {
+  _rustLanguageService: AtomLanguageService<LanguageService>;
+  _buckTaskRunnerService: ?BuckTaskRunnerService;
```
mostafaeweda: we don't seem to need this to be an instance variable, right?

pkg/nuclide-rust/lib/main.js
```diff
@@ -0,0 +1,58 @@
...
+    atom.notifications.addInfo(DISCLAIMER);
+
+    this._rustLanguageService = createRustLanguageService();
+    this._rustLanguageService.activate();
+
+    this._subscriptions = new UniversalDisposable(this._rustLanguageService);
+  }
+
+  consumeBuckTaskRunner(service: BuckTaskRunnerService): IDisposable {
+    service.onDidCompleteTask(task =>
```
mostafaeweda: that's a non disposed registration, maybe this function should just:

`return service.onDidCompleteTask...`

pkg/nuclide-rust/lib/BuckIntegration.js
Xanewok: Done in https://github.com/facebook/nuclide/pull/1606/commits/8d4772f9ecda99b132d0f4b8eced16b9124455af. Forgot to remove the comment, sorry!

pkg/nuclide-rust/lib/BuckIntegration.js
Xanewok: I understood that MultiLspLanguageService spawn separate LSP server per appropriate 'top-level' project config files like .hhconfig and that calling this seems like a good way to retrieve a handle to appropriate LSP server that's responsible for the fileUri we're building.

However, I did assume there's going to be a single Buck root; I'll see if it works with multiple ones.

pkg/nuclide-rust/lib/BuckIntegration.js
Xanewok: It depends; using `cargo check` as example, complete check build can even take a minute or two; rechecking the leaf only shouldn't take longer than a couple of seconds. Third party deps are vendored and pre-compiled from what I understand, so this should also further cut down compilation time. @jsgf would know more about how long exactly these `#save-analysis` builds take in the Buck environment.

It would certainly be good to report the progress - do you know if it's possible to pass a callback or report the progress gracefully using the Buck API here?

Xanewok: Changed this to display a busy spinner indicator while the save-analysis build executes.

pkg/nuclide-rust/lib/BuckIntegration.js
Xanewok: @mostafaeweda sorry it took so long! Just tested this with multiple buck roots, using https://github.com/Xanewok/rust-buck-skeleton and building both `//server:server` and `nested//:nested`. This correctly retrieved the LSP server at root dir and nested/, respectively, so this looks good so far =)

Xanewok:
@mostafaeweda @jsgf Rebased and pushed two changes: one being displaying busy indicator while the Buck builds the save-analysis build and the other one working around a Buck bug (https://github.com/facebook/nuclide/pull/1606/commits/9f64faf40f56e04178c0af7feef7a298a3ae98bf), where outputs are relative to the cell, even when these are executed from the outer Buck root.

Does this still need more work or can we land it as the 1st increment?

pkg/nuclide-rust/lib/BuckIntegration.js
```diff
@@ -0,0 +1,137 @@
...
+  const buildTarget = normalizeNameForBuckQuery(task.buildTarget);
+
+  // Output is relative to Buck root but the built target may be managed by a
+  // Buck cell (nested Buck root).
+  // Here, Buck returns input paths relative to the possible cell, but the build
+  // file always relative to the current Buck root. Because of that, we use the
+  // build file path to determine the possible Buck cell root to which the
+  // inputs are relative to.
+  // FIXME: This is a bug in Buck, only query for files when the output is fixed.
+  const [relativeBuildFile, files] = await Promise.all([
```
mostafaeweda: Before issuing these 2 potentially-expensive buck calls, could we verify that's actually a Rust target? (maybe using `task.buildRuleType.type`?

Xanewok:
@jsgf signed the CLA here, just in case.

pkg/nuclide-rust/lib/BuckIntegration.js
Xanewok: @mostafaeweda my bad, I misunderstood what you were saying - I thought the gist was there being 2 calls and not checking before issuing them. Sorry! Will fix that in a moment.

Xanewok: Done [here](https://github.com/facebook/nuclide/pull/1606/commits/7b9c326d1646262b6270e2e5c94353b96c83d152#diff-d7cc36e67cb07d1c82a31ad33a2544efR38).

PR #1626, opened 2018-10-03
added show on cursor over for datatip

udnaan:
This pull request adds datatip on cursor while using keyboard. The setting can be turned on by going into settings.

This is the same [pull request](https://github.com/facebook-atom/atom-ide-ui/pull/307)


kcaze:
This doesn't seem right. I put some logging in `_startFetching()` and confirmed that if this is turned on, it'll try to fetch datatips on every single keystroke.

This feels like it should at least do `_startFetchingDebounce()` instead of `_startFetching()`?

udnaan:
I switched to startFetchingDebounce but in a local test, this doesn't do what it is suppose to do. I will have to stick with the startFetching unless someone has any insights.

kimgronqvist:
I would also like this and would be happy to test and debug any issues that needs to be fixed.

